PR #4, opened 2023-10-20
[FEATURE] Add monorepo validation setup

jaapio:
To ensure we keep things working, add some validation on the repository structure.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,0 +1,29 @@
+name: "Main"
+
+on: # yamllint disable-line rule:truthy
+  pull_request: null
+  push:
+    branches:
+      - "main"
+
+jobs:
+  monorepo-validate:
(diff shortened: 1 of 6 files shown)


.github/workflows/main.yaml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "8.2"
```
brotkrueml, line 20: Should we use PHP 8.1, as this version is the minimum requirement of TYPO3? Or is this the minimum version of the parser?

jaapio: We can use php 8.1 here

brotkrueml (approved):

PR #6, opened 2023-10-20
[TASK] Use the vendor folder for Composer packages

brotkrueml:


Dockerfile
```diff
@@ -10,9 +10,9 @@ FROM php:8.2-cli-alpine
 COPY . /opt/guides
 WORKDIR /opt/guides
 
-COPY --from=Builder /opt/guides/.Build /opt/guides/.Build
-RUN cp guides.xml .Build/guides.xml
+COPY --from=Builder /opt/guides/vendor /opt/guides/vendor
+RUN cp guides.xml guides.xml
```
brotkrueml, line 14: @jaapio Is the the correct path then? `.Build/guides.xml` -> `guides.xml`

linawolf (approved):

Dockerfile
linawolf: yes that is correct

linawolf: however we do not have to copy it then. The guides always searches in the parent directory of vendor

linawolf, line 14:
```suggestion
```
This line does not make sense, lets remove it

jaapio: Yeah, we can simply remove this cp command

brotkrueml: Done

PR #7, opened 2023-10-21
[TASK] Introduce TYPO3 CS for PHP

linawolf:
I think we should also run this in the Github Actions but I am not sure how

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,0 +1,12 @@
+<?php
+
+$config = \TYPO3\CodingStandards\CsFixerConfig::create();
+$config
+    ->getFinder()->in(__DIR__)
+    ->exclude([
+        'Documentation',
+        'docs',
+    ])
+;
+
+return $config;
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -10,6 +10,27 @@ env:
   DEFAULT_PHP_VERSION: "8.1"
 
 jobs:
+  quality:
+    name: Quality
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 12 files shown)


brotkrueml:
> I think we should also run this in the Github Actions but I am not sure how

Will have a look shortly.

.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+$config = \TYPO3\CodingStandards\CsFixerConfig::create();
+$config
+    ->getFinder()->in(__DIR__)
+    ->exclude([
+        'Documentation',
```
brotkrueml, line 7: What is the difference of the both directories? "Documentation" is not available yet.

linawolf: I always put the stuff I want to test into Documentation

brotkrueml: Then, should be the stuff from "Documentation" be code-styled?

And what is the purpose of "docs"? Actually, there is only one rst file. Should it be delivered to the users?

linawolf: The docs was just added by @jaapio and explains how you can use the mono-repository. So I guess it should be delivered to the user if the complete reposiory is delivered to them?

linawolf: And no, I wouldn't want an external script to code-style my test Data as it comes from a different source, might have different PHP versions etc

brotkrueml: Aaaah, it is an "exclude", sorry, just looked at the folder names....

brotkrueml (approved):

PR #8, opened 2023-10-21
[TASK] Add .gitattributes file

brotkrueml:
The files/folders defined in the file ("export-ignore") are not delivered when requesting a version from packagist.


linawolf:
we should add .php-cs-fixer.dist.php after merging #7

brotkrueml:
> we should add .php-cs-fixer.dist.php after merging #7

Yup, that's why I set it to "draft"

brotkrueml:
This is the wrong place. Should be added into the distinct packages, I assume.

PR #10, opened 2023-10-22
[TASK] Rename "Documentation" folder to fixtures-local

linawolf:
This makes clear it does not contain information about the project but test-data

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 /output/
 /vendor/
-/Documentation/
+/fixtures-local/
 /.php-cs-fixer.cache
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -4,8 +4,8 @@ $config = \TYPO3\CodingStandards\CsFixerConfig::create();
 $config
     ->getFinder()->in(__DIR__)
     ->exclude([
-        'Documentation',
         'docs',
+        'fixtures-local',
     ])
 ;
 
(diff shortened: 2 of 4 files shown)


brotkrueml:
> This makes clear it does not contain information about the project but test-data

I assume, running the command will error when the folder is not available. I suggest to create the folder with an empty `.gitkeep` file in it (or a readme file explaining what the purpose of this folder is).

linawolf:
@brotkrueml Well the README states that you should copy your documentation to be rendered into that directory if you want to use the package with ddev?

brotkrueml:
You mean this one?

    cp -R /path/to/my/docs/Documentation fixtures-local

Currently, I don't get it. Why do I need ddev and why do I have to copy my extension's Documentation folder into this project to get rendered?

linawolf:
You don't have to have ddev. But if you want to use ddev - which is an option that many ppl like you can use it. However ddev has restrictions that you cannot access files outside the main folder of the project. Therefore if you want to render test data that way you have to move it into your project and put it somewhere :tm: If you dont want to do that use docker and run it whereever you like or install PHP 8.2 on your local machine (something I for example didnt get working on my WSL yet)

brotkrueml:
Then the workflow and the purpose of the directory should be documented in the README (not part of this PR though).

But you can use, for example, a setup like here with a symlink and a `docker-compose.volumes.yaml`:
https://brot.krue.ml/using-ddev-for-development-of-typo3-extensions/
This way, you don't have to copy docs after updates around. But this is just for the records.

brotkrueml (approved):

PR #11, opened 2023-10-22
[FEATURE] Introduce PHP stan

linawolf:
Put existings errors in the baseline for now, they can be analyzed and solved in a follow up

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -29,10 +29,10 @@ jobs:
           dependency-versions: "locked"
 
       - name: CGL
-        run: composer run check:php:cs
+        run: composer run check:php:php-cs-fixer
 
       - name: PHPSTAN
-        run: composer run check:php:stan
+        run: composer run check:php:phpstan
 
   monorepo-validate:
     runs-on: "ubuntu-latest"
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -31,6 +31,9 @@ jobs:
       - name: CGL
         run: composer run check:php:cs
 
+      - name: PHPSTAN
+        run: composer run check:php:stan
+
   monorepo-validate:
     runs-on: "ubuntu-latest"
     steps:
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,9 @@
     },
     "require-dev": {
         "ergebnis/composer-normalize": "^2.30",
+        "phpstan/extension-installer": "^1.3",
+        "phpstan/phpstan": "^1.10",
+        "phpstan/phpstan-strict-rules": "^1.5",
         "symplify/monorepo-builder": "^11.2",
         "typo3/coding-standards": "^0.7.1"
     },
@@ -25,12 +28,18 @@
     },
     "config": {
         "allow-plugins": {
-            "ergebnis/composer-normalize": true
+            "ergebnis/composer-normalize": true,
+            "phpstan/extension-installer": true
         },
         "sort-packages": true
     },
     "scripts": {
+        "check:php": [
+            "composer check:php:cs",
+            "composer check:php:stan"
+        ],
         "check:php:cs": "vendor/bin/php-cs-fixer check",
+        "check:php:stan": "vendor/bin/phpstan --configuration=phpstan.neon",
         "fix:php:cs": "vendor/bin/php-cs-fixer fix",
         "monorepo:merge": "vendor/bin/monorepo-builder merge",
         "monorepo:validate": "vendor/bin/monorepo-builder validate",
--- a/composer.json
+++ b/composer.json
@@ -35,11 +35,11 @@
     },
     "scripts": {
         "check:php": [
-            "composer check:php:cs",
-            "composer check:php:stan"
+            "composer check:php:php-cs-fixer",
+            "composer check:php:phpstan"
         ],
-        "check:php:cs": "vendor/bin/php-cs-fixer check",
-        "check:php:stan": "vendor/bin/phpstan --configuration=phpstan.neon",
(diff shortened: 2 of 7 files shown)


phpstan.neon
```diff
@@ -0,0 +1,13 @@
...
+  level: max
+  strictRules:
+    allRules: false
+    strictCalls: true
+    requireParentConstructorCall: true
+  inferPrivatePropertyTypeFromConstructor: true
+  treatPhpDocTypesAsCertain: false
+  paths:
+    - packages/typo3-docs-theme/src
+  tmpDir: .build/phpstan/
```
brotkrueml, line 13: What is the benefit of defining `tmpDir` explicitly? Default is `/tmp/phpstan`.
Using the default one: there is no need to introduce a new folder `.build`

phpstan.neon
```diff
@@ -0,0 +1,13 @@
+includes:
+  - phpstan-baseline.neon
+parameters:
+  level: max
```
brotkrueml, line 4:
```suggestion
  phpVersion: 80100
  level: max
```

I suggest to run phpstan always in PHP 8.1, no matter which PHP version is used. I would prefer this, is a mostly run that on my Linux box directly instead using docker.

composer.json
```diff
@@ -3,12 +3,17 @@
     "license": "OPL-1.0",
     "require": {
         "php": "^8.1",
+        "jangregor/phpstan-prophecy": "^1.0",
```
brotkrueml, line 6: Why this one? We don't use prophecy, do we?

composer.json
```diff
@@ -3,12 +3,17 @@
...
         "phpdocumentor/guides-cli": "dev-main",
         "phpdocumentor/guides-theme-bootstrap": "dev-main",
+        "phpstan/extension-installer": "^1.3",
         "symfony/http-client": "^6.3"
     },
     "require-dev": {
         "ergebnis/composer-normalize": "^2.30",
+        "phpstan/phpstan": "^1.10",
+        "phpstan/phpstan-strict-rules": "^1.5",
+        "phpstan/phpstan-webmozart-assert": "^1.2",
```
brotkrueml, line 16: Why this one, do we use webmozart/assert?

phpstan.neon
linawolf: @jaapio I copied that from the guides. is there an advantage to this or should we use the default as suggested by @brotkrueml ?

composer.json
linawolf: true, removed it until we need it

composer.json
linawolf: removed it until we need it

phpstan.neon
linawolf: I made this a serparate commit so we can preserve why the descision was made

phpstan.neon
linawolf: In DDEV the phpstan cache does not get stored in my file system then. Not sure if this is an advantage or disadvantage?

jaapio: the phpstan cache will speed up the analysis, if we configure this correctly it will work in ddev and docker, if you have php natively installed you will not have an advantage rather than that the cache will not be cleared when your system cleans the /tmp dir. 

I think it's good to configure this.

.github/workflows/main.yaml
```diff
@@ -31,6 +31,9 @@ jobs:
       - name: CGL
         run: composer run check:php:cs
 
+      - name: PHPSTAN
+        run: composer run check:php:stan
```
jaapio, line 35: this seems like a very odd naming to me... is this some typo3 standard?

phpstan.neon
linawolf: would it make sense to call the directory .cache instead of .build to make sure it can and should be cleared if there are problems with cache?

brotkrueml: That would be much better, as the purpose of this folder is now clear. I suggest to use `.cache/phpstan`, maybe we have other dev caches later.

.github/workflows/main.yaml
linawolf: what exactly is odd about this? How would you call that? I have no idea what that name means I only tried to copy from the cgl check

phpstan.neon
linawolf: I think, we could also move the cs-fixer cache there and maybe the phpunit cache?

.github/workflows/main.yaml
brotkrueml: I think he means the "check:php:stan" instead of using "check:php:phpstan". The tool is not named "Stan".

linawolf: we do the same for php:cs?

jaapio: Ah I get the confusion now :-P 

`php:cs` => `php:code-style`
'php:stan => `php:static-analysis` 

I thought this was some kind of odd naming convention to convert all tool names to names without `-` (dash)  or camel casing.

brotkrueml: But "stan" is no package. it is called "phpstan", so "check:php:phpstan" would be better. And I never thought of the name that that would mean "STatic ANalysis". And I doubt we would switch to psalm (or phan) in the next time. But just my 2 cents. I do not block it.

linawolf: we could use 
php:cs => php:code-style
php:stan => php:static-analysis

That would not be bound to one tool and maybe easier to understand for newbies?

jaapio: This isn't blocking for me either.

But it makes sense to keep the naming inline with the commands used to run this locally

linawolf: Ok so I renamed them for the tools names

jaapio (approved):

PR #13, opened 2023-10-22
[TASK] Do not support the uml directive

linawolf:
It is immensely complex and was never adopted by the community. We do not use it in the official docs.


brotkrueml:
> It is immensely complex and was never adopted by the community. We do not use it in the official docs.

That's sad, I use it and like it very much:
https://docs.typo3.org/p/brotkrueml/schema/2.11/en-us/Developer/Api.html#the-model-in-depth

linawolf:
Ok so this cannot be done with graphviz? If it is really needed (I was doing search and I think even asking in the channel) then we have to find a way to support it

brotkrueml:
> Ok so this cannot be done with graphviz? If it is really needed (I was doing search and I think even asking in the channel) then we have to find a way to support it

Don't know about it, will try.

linawolf:
Ah, sorry, my bad. it is graphviz we want to drop and uml (plantuml) we want to keep

PR #14, opened 2023-10-22
[FEATURE] Install guides-graphs for plantUML support

linawolf:


composer.json
```diff
@@ -24,7 +25,8 @@
     },
     "config": {
         "allow-plugins": {
-            "ergebnis/composer-normalize": true
+            "ergebnis/composer-normalize": true,
+            "phpstan/extension-installer": true
```
brotkrueml, line 29: This one is not part of the PR?

brotkrueml (changes requested):

composer.json
linawolf: removed it

brotkrueml (approved):

PR #16, opened 2023-10-22
[FEATURE] Introduce integration tests

linawolf:
Those are run by phpunit

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
         "phpstan/extension-installer": "^1.3",
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-strict-rules": "^1.5",
+        "phpunit/phpunit": "^10.4",
         "symplify/monorepo-builder": "^11.2",
         "typo3/coding-standards": "^0.7.1"
     },
@@ -28,6 +29,11 @@
             "T3Docs\\Typo3DocsTheme\\": "packages/typo3-docs-theme/src/"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "T3Docs\\Typo3DocsTheme\\": "tests/"
+        }
+    },
     "config": {
         "allow-plugins": {
             "ergebnis/composer-normalize": true,
@@ -36,6 +42,7 @@
         "sort-packages": true
     },
     "scripts": {
+        "build:phar": "tools/build-phar.sh",
         "check:php": [
             "composer check:php:php-cs-fixer",
             "composer check:php:phpstan"
@@ -46,6 +53,6 @@
         "monorepo:merge": "vendor/bin/monorepo-builder merge",
         "monorepo:validate": "vendor/bin/monorepo-builder validate",
         "render": "vendor/bin/guides fixtures-local output --theme=typo3docs",
-        "build:phar": "tools/build-phar.sh"
+        "test:integration": "vendor/bin/phpunit --testsuite=integration"
     }
 }
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<phpunit
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.1/phpunit.xsd"
+        bootstrap="vendor/autoload.php"
+        colors="true"
+        beStrictAboutTestsThatDoNotTestAnything="true"
+        beStrictAboutOutputDuringTests="true"
+        beStrictAboutChangesToGlobalState="true"
+        failOnNotice="true"
+        failOnWarning="true"
+        displayDetailsOnTestsThatTriggerDeprecations="true"
+        displayDetailsOnTestsThatTriggerNotices="true"
+        displayDetailsOnTestsThatTriggerWarnings="true"
+        displayDetailsOnTestsThatTriggerErrors="true"
+        cacheDirectory=".cache/.phpunit.cache"
+        requireCoverageMetadata="false"
+>
+    <testsuites>
+        <testsuite name="integration">
+            <directory>./tests/Integration/</directory>
+        </testsuite>
+    </testsuites>
+    <coverage/>
+    <source>
+        <include>
+            <directory suffix=".php">packages/**/src</directory>
+        </include>
+    </source>
+</phpunit>
--- a/tests/ApplicationTestCase.php
+++ b/tests/ApplicationTestCase.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace T3Docs\Typo3DocsTheme;
+
+use phpDocumentor\Guides\Bootstrap\DependencyInjection\BootstrapExtension;
+use phpDocumentor\Guides\Cli\DependencyInjection\ApplicationExtension;
+use phpDocumentor\Guides\Cli\DependencyInjection\ContainerFactory;
+use phpDocumentor\Guides\DependencyInjection\TestExtension;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\DependencyInjection\Container;
+use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
+use T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension;
+
+abstract class ApplicationTestCase extends TestCase
+{
+    private Container|null $container = null;
+
+    public function getContainer(): Container
+    {
+        if ($this->container === null) {
+            $this->prepareContainer();
+        }
+
+        return $this->container;
+    }
+
+    /**
+     * @param array<string, array<mixed>> $configuration
+     * @param ExtensionInterface $extraExtensions
+     *
+     * @phpstan-assert Container $this->container
+     */
+    protected function prepareContainer(string|null $configurationFile = null, array $configuration = [], array $extraExtensions = []): void
+    {
+        $containerFactory = new ContainerFactory([
+            new ApplicationExtension(),
+            new TestExtension(),
+            new BootstrapExtension(),
+            new Typo3DocsThemeExtension(),
+            ...$extraExtensions,
+        ]);
+
+        foreach ($configuration as $extension => $extensionConfig) {
+            $containerFactory->loadExtensionConfig($extension, $extensionConfig);
+        }
+
+        if ($configurationFile !== null) {
+            $containerFactory->addConfigFile($configurationFile);
+        }
+
+        $this->container = $containerFactory->create(dirname(__DIR__) . '/vendor');
+    }
+}
--- a/tests/Integration/IntegrationTest.php
+++ b/tests/Integration/IntegrationTest.php
@@ -1,0 +1,200 @@
+<?php
+
+declare(strict_types=1);
+
+namespace T3Docs\Typo3DocsTheme\Integration;
+
+use function array_filter;
+use function array_merge;
+use function array_walk;
+use function assert;
+use function escapeshellarg;
+use function explode;
+use function file_exists;
+use function file_get_contents;
+use function implode;
+
+use const LC_ALL;
+
+use phpDocumentor\Guides\Cli\Command\Run;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\ExpectationFailedException;
+
+use function setlocale;
+use function str_ends_with;
+use function str_replace;
+
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+use Symfony\Component\Finder\Finder as SymfonyFinder;
+
+use function system;
+
+use T3Docs\Typo3DocsTheme\ApplicationTestCase;
+
+use function trim;
+
+final class IntegrationTest extends ApplicationTestCase
+{
+    protected function setUp(): void
+    {
+        setlocale(LC_ALL, 'en_US.utf8');
+    }
+
+    /** @param list<string> $compareFiles */
+    #[DataProvider('getTestsForDirectoryTest')]
+    public function testHtmlIntegration(
+        string $inputPath,
+        string $expectedPath,
+        string $outputPath,
+        array $compareFiles,
+    ): void {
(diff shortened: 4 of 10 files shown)


composer.json
```diff
@@ -23,13 +24,19 @@
             "T3Docs\\Typo3DocsTheme\\": "packages/typo3-docs-theme/src/"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "T3Docs\\Typo3DocsTheme\\": "tests/"
```
brotkrueml, line 29: Why tests folder not under "packages/typo3-docs-theme/tests/". I assume, there may be some other packages in the future? That's why we have a monorepo?

tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,206 @@
...
+use function file_get_contents;
+use function implode;
+
+use const LC_ALL;
+
+use phpDocumentor\Guides\Cli\Command\Run;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\ExpectationFailedException;
+
+use function setlocale;
```
brotkrueml, line 23: Do we really need those "use function" thingy? It clutters the "use" section IMHO. Either avoid it or add a `\` before the functions when using it to signal it is in the global namespace.

tests/integration/IntegrationTest.php
```diff
@@ -0,0 +1,205 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\Integration;
+
+use function array_filter;
```
brotkrueml, line 7: As mentioned above, we should avoid "use function".

tests/Integration/tests/index/expected/objects.inv.json
```diff
@@ -0,0 +1,18 @@
+{
```
brotkrueml, line 1: Why is this file part of the integration tests? It is not checked anywhere?

tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,206 @@
...
+            $value = trim($value);
+        });
+        $contentArray = array_filter($contentArray, static function ($value) {
+            return $value !== '';
+        });
+
+        return implode("\n", $contentArray);
+    }
+
+    /** @return mixed[] */
```
brotkrueml, line 158: I don't think this is useful. Do we really want to have this on tests, then we should annotate the proper return type. Also below.

tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,206 @@
...
+    public static function getTestsForLatex(): array
+    {
+        return self::getTestsForDirectory(__DIR__ . '/tests-latex');
+    }
+
+    /** @return mixed[] */
+    private static function getTestsForDirectory(string $directory): array
+    {
+        $finder = new SymfonyFinder();
```
brotkrueml, line 173: Please add symfony/finder as dev dependency. This makes it explicit, that the package is used - and in which version.

phpunit.xml.dist
```diff
@@ -0,0 +1,30 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.1/phpunit.xsd"
+        bootstrap="vendor/autoload.php"
+        colors="true"
+        beStrictAboutTestsThatDoNotTestAnything="true"
+        beStrictAboutOutputDuringTests="true"
+        beStrictAboutChangesToGlobalState="true"
+        failOnNotice="true"
+        failOnWarning="true"
+        displayDetailsOnTestsThatTriggerNotices="true"
```
brotkrueml, line 13: I suggest to add also "displayDetailsOnTestsThatTriggerDeprecations" to get early hints on deprecations.

phpunit.xml.dist
```diff
@@ -0,0 +1,30 @@
...
+        displayDetailsOnTestsThatTriggerWarnings="true"
+        displayDetailsOnTestsThatTriggerErrors="true"
+        cacheDirectory=".cache/.phpunit.cache"
```
brotkrueml, line 16: Can we then just use ".cache/phpunit"? No need to make it hidden with "." in the ".cache" folder.

tests/ApplicationTestCase.php
```diff
@@ -0,0 +1,55 @@
...
+use phpDocumentor\Guides\Bootstrap\DependencyInjection\BootstrapExtension;
+use phpDocumentor\Guides\Cli\DependencyInjection\ApplicationExtension;
+use phpDocumentor\Guides\Cli\DependencyInjection\ContainerFactory;
+use phpDocumentor\Guides\DependencyInjection\TestExtension;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\DependencyInjection\Container;
+use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
+use T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension;
+
+class ApplicationTestCase extends TestCase
```
brotkrueml, line 16: Should we make this class "abstract" (like TestCase is also abstract)? So it is clear that is is meant for extending?

tests/integration/IntegrationTest.php
```diff
@@ -0,0 +1,205 @@
...
+
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+use Symfony\Component\Finder\Finder as SymfonyFinder;
+
+use function system;
+
+use function trim;
+
+class IntegrationTest extends ApplicationTestCase
```
brotkrueml, line 37: Should we make this class "abstract" (like TestCase is also abstract)? So it is clear that is is meant for extending?

tests/integration/IntegrationTest.php
```diff
@@ -0,0 +1,205 @@
...
+{
+    protected function setUp(): void
+    {
+        setlocale(LC_ALL, 'en_US.utf8');
+    }
+
+    /** @param string[] $compareFiles */
```
brotkrueml, line 44:
```suggestion
    /** @param list<string> $compareFiles */
```

This makes it clear that there are no "holes" as keys (like 0, 2, 3 where 1 is missing).

tests/Integration/tests/index/expected/index.html
```diff
@@ -0,0 +1,168 @@
+<!DOCTYPE html>
```
brotkrueml, line 1: Do we really need the whole HTML structure or just a part of it? So checking in the texts if the expected output is part of the whole actual HTML structure? This makes maintaining more easy, when we change the head output, for example.

composer.json
linawolf: This is an integration test that does not only test our theme but the complete integration with other installed packages etc

tests/Integration/tests/index/expected/objects.inv.json
linawolf: yes it is. It is the expected output of the file

tests/Integration/tests/index/expected/index.html
linawolf: We treat the output like a baseline. You are alerted about any change and if the change is fine you can just copy the accepted output back into the integration. In the guides this strategy worked fine for us. Each Commit which causes a change in the output will then also be in the Git-History of that file.

tests/Integration/IntegrationTest.php
linawolf: I believe this is the doctrine code-style. But honestly I wouldnt want to change these files where not absolutely nessesary as it will make it harder to cherry pick changes from https://github.com/phpDocumentor/guides/blob/main/tests/ApplicationTestCase.php and https://github.com/phpDocumentor/guides/blob/main/tests/Integration/IntegrationTest.php into this test. I would suggest that you make a PR to change those things that make sense in the original files instead?

tests/ApplicationTestCase.php
linawolf: we could but we should do that with the original test case

tests/integration/IntegrationTest.php
linawolf: It should first be changed in the guides then

phpunit.xml.dist
linawolf: I think ".phpunit.cache" is like the standard name of phpunit cache. Why should we rename it and confuse ppl?

composer.json
brotkrueml: Ack

tests/Integration/IntegrationTest.php
brotkrueml: Yes, can be done in a separate PR. IMHO it should be a setting in php-cs-fixer.

phpunit.xml.dist
brotkrueml: Which people care about a phpunit cache directory, but okay ...

tests/ApplicationTestCase.php
brotkrueml: So the ApplicationTestCase can run as a test?

tests/integration/IntegrationTest.php
brotkrueml: Sorry, mixed it up.

tests/integration/IntegrationTest.php
brotkrueml: Why? The $compareFiles comes from the data provider "getTestsForDirectoryTest()" in the same class?

tests/Integration/tests/index/expected/index.html
brotkrueml: Okay, may be error-prone, if you have many changes in the output and some erroneous one slipps through.

tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,206 @@
...
+
+        self::assertFalse($skip, 'Test passes while marked as SKIP.');
+    }
+
+    /**
+     * Asserts that each line of the expected file is contained in actual
+     *
+     * @throws ExpectationFailedException
+     */
+    public static function assertFileContainsLines(string $expected, string $actual): void
```
brotkrueml, line 119: Why public?

tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
+    /**
+     * Asserts that the contents of one file is equal to the contents of another
+     * file. It ignores empty lines and whitespace at the start and end of each line
+     *
+     * @throws ExpectationFailedException
+     */
+    public static function assertFileEqualsTrimmed(string $expected, string $actual, string $message = ''): void
```
brotkrueml, line 137: Why public?

tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,206 @@
...
+    {
+        self::assertFileExists($expected, $message);
+        self::assertFileExists($actual, $message);
+
+        self::assertEquals(self::getTrimmedFileContent($expected), self::getTrimmedFileContent($actual), $message);
+    }
+
+    public static function getTrimmedFileContent(string $file): string
```
brotkrueml, line 145: Why public?

tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,206 @@
...
+        });
+        $contentArray = array_filter($contentArray, static function ($value) {
+            return $value !== '';
+        });
+
+        return implode("\n", $contentArray);
+    }
+
+    /** @return mixed[] */
+    public static function getTestsForDirectoryTest(): array
```
brotkrueml, line 159: Why public?

tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,206 @@
...
+    {
+        return self::getTestsForDirectory(__DIR__ . '/tests');
+    }
+
+    /** @return mixed[] */
+    public static function getTestsForLatex(): array
```
brotkrueml, line 165: Why public?

tests/Integration/IntegrationTest.php
linawolf: Because phpunit expects data providers to be public even if the are used within the same file

tests/Integration/IntegrationTest.php
linawolf: changed in all locations

tests/ApplicationTestCase.php
linawolf: no, you are right, made it abstract

tests/Integration/IntegrationTest.php
linawolf: doesn't need to be public as it is no data provider

tests/Integration/IntegrationTest.php
linawolf: Removed the data provider as we do not use it

tests/integration/IntegrationTest.php
linawolf: adjusted

tests/Integration/IntegrationTest.php
linawolf: Symfony/finder is already in the require section

phpunit.xml.dist
linawolf: Added it

phpunit.xml.dist
linawolf: I actually had problems where I needed to delete it to get rid of problems in the guides several times...

tests/Integration/IntegrationTest.php
brotkrueml: Ack, you are right.

linawolf:
@brotkrueml any other changes required or can we merge this?

composer.json
```diff
@@ -41,6 +47,7 @@
         ],
         "check:php:php-cs-fixer": "vendor/bin/php-cs-fixer check",
         "check:php:phpstan": "vendor/bin/phpstan --configuration=phpstan.neon",
+        "check:integration": "vendor/bin/phpunit --testsuite=integration",
```
brotkrueml, line 50: I would prefer `tests:integration`.

linawolf: As the others are all verbs, how about test:integration

brotkrueml: Go for it

brotkrueml (approved):

PR #17, opened 2023-10-22
[FEATURE] Introduce a shell script for the rendering order

linawolf:
Currently if there is a Index.rst in folder Documentation it is rendered. If not in that order README.rst or README.md are rendered.

Currently the markdown extension is not installed yet, will do that in a follow up

### DIFF
--- a/render.sh
+++ b/render.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+documentation_folder="Documentation"
+index_file="$documentation_folder/Index.rst"
+readme_rst_file="README.rst"
+readme_md_file="README.md"
+output_folder="output"
+
+
+# Ensure the output folder exists
+mkdir -p "$output_folder"
+
+if [ -f "$index_file" ]; then
+    # Index.rst exists, render from there
+    vendor/bin/guides  "$documentation_folder" "$output_folder"
+elif [ -f "$readme_rst_file" ]; then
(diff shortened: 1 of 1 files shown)


render.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+documentation_folder="Documentation"
+index_file="$documentation_folder/Index.rst"
+readme_rst_file="README.rst"
+readme_md_file="README.md"
+output_folder="output"
```
brotkrueml, line 7: Currently, I don't like the folder name "output" at all. What output? Of a test? Of a custom script from an extension? OF the docs? Can we keep the well-know Documentation-GENERATED-temp?

brotkrueml (commented):
In which context should this be called? Please document it in the readme file.

render.sh
linawolf: Yes, why not

linawolf:
> In which context should this be called? Please document it in the readme file.

THIS is a first POV not sure yet if this would be the  way to go

jaapio:
I would like to check if and how we can move this behaviour into the main app, via an extension. If that removes the need for extra scripts, it might be worth it.

linawolf:
Converting this to draft for now

jaapio:
With #20 merged we do not need this anymore?

linawolf:
Right

PR #18, opened 2023-10-26
[TASK] build phar

jaapio:
This adds a script to build a phar file from this repository.
Some extra scripting is required due a limitation in box regaring
composer plugins. The rest is just plain box building.

### DIFF
--- a/.github/workflows/phar.yaml
+++ b/.github/workflows/phar.yaml
@@ -7,6 +7,7 @@ on:
     types:
       - published
 
+name: Phar
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


linawolf:
@jaapio so when I try to run the new script tools/build-phar.sh it changes the composer.json and composer.lock is this intended?

linawolf:
> @jaapio so when I try to run the new script tools/build-phar.sh it changes the composer.json and composer.lock is this intended?

Ah ok it is written in the README, never mind

linawolf (approved):
So as far as I can tell it looks fine and does not seem to be to hard to maintain

PR #20, opened 2023-10-26
[FEATURE] Add automated input detection

jaapio:
By decorating the run command we can add extra functionality to the default command of guides. In this case I added the input guessing logic to find the correct input format or single file readme documentation.

### DIFF
--- a/packages/typo3-guides-extension/src/Command/RunDecorator.php
+++ b/packages/typo3-guides-extension/src/Command/RunDecorator.php
@@ -1,0 +1,150 @@
+<?php
+
+namespace T3Docs\GuidesExtension\Command;
+
+use phpDocumentor\Guides\Cli\Command\Run;
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Input\InputDefinition;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class RunDecorator extends Command
+{
+    private const DEFAULT_OUTPUT_DIRECTORY = 'Documentation-GENERATED-temp';
+
+    private const INDEX_FILE_NAMES = [
+        'Index.rst' => 'rst',
+        'index.rst' => 'rst',
+        'Index.md' => 'md',
+        'index.md' => 'md',
+    ];
+    private const FALLBACK_FILE_NAMES = [
+        'README.rst' => 'rst',
+        'README.md' => 'md',
+    ];
+
+    private Run $innerCommand;
+    public function __construct(Run $innerCommand)
+    {
+        parent::__construct($innerCommand->getName());
+        $this->innerCommand = $innerCommand;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $options = [];
+        foreach ($input->getOptions() as $option => $value) {
+            if ($value === null) {
+                continue;
+            }
+
+            $options['--' . $option] = $value;
+        }
+
+        $arguments = $input->getArguments();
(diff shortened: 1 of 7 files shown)


jaapio:
@linawolf I think this makes it easier for you to maintain the way the input detection is handled. It's plain php, this makes it possible to run the phar file without knowledge about extra bash scripts. Both the docker image and phar file will work in exactly the same way this way. No extra scripts to be loaded or environments to setup. 

It's a bit of a hack, because Symfony does not really provide a way to hook into the commands, but the way I implemented this should be stable.  

The reason to introduce a new extension for this was to keep the theme and real extension separated, of course we could bring this into the already existing extension, please let me know what you prefer.

packages/typo3-guides-extension/src/Command/RunDecorator.php
```diff
@@ -0,0 +1,150 @@
...
+    private function guessInput(OutputInterface $output): array
+    {
+        $currentDirectory = getcwd();
+        if ($currentDirectory === false) {
+            return [];
+        }
+
+        $inputDirectory = $currentDirectory . '/Documentation';
+
+        if (file_exists($inputDirectory)) {
```
brotkrueml, line 117:
```suggestion
        if (is_dir($inputDirectory)) {
```

packages/typo3-guides-extension/src/Command/RunDecorator.php
```diff
@@ -0,0 +1,150 @@
...
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Input\InputDefinition;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class RunDecorator extends Command
+{
+    private const DEFAULT_OUTPUT_DIRECTORY = 'Documentation-GENERATED-temp';
+
+    private const INDEX_FILE_NAMES = [
```
brotkrueml, line 17: What about using one array like

    [
        'Index.rst',
        'index.rst',
        'Index.md',
        'index.md',
    ]
    
This way it is clearer in which order the index files are searched for. And we can omit one foreach loop in `guessInput()`.

packages/typo3-guides-extension/src/Command/RunDecorator.php
```diff
@@ -0,0 +1,150 @@
...
+                            'input' => $inputDirectory,
+                            '--input-format' => $extension,
+                        ];
+                    }
+                }
+            }
+        }
+
+        if ($output->isVerbose()) {
+            $output->writeln('Index documetation file not found, trying README.rst or README.md');
```
brotkrueml, line 135:
```suggestion
            $output->writeln('Index documentation file not found, trying README.rst or README.md');
```

Maybe add a new class constant FALLBACK_FILE_NAMES with README.rst and README.md as content. This way we have one place and can use implode on that array for the message. Imagine we allow later also Readme.md and Readme.rst. Just adjust the array and everything is fine.

packages/typo3-guides-extension/src/Command/RunDecorator.php
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        foreach (self::EXTENSIONS as $extension) {
```
brotkrueml, line 138: When introducing a FALLBACK_FILE_NAMES class constant constant this can be streamlined.

.gitignore
```diff
@@ -1,5 +1,5 @@
+*GENERATED*
 /output/
 /vendor/
-/fixtures-local/
```
linawolf, line 3: *GENERATED* is a replacement for output, not for fixtures-local

brotkrueml: Done

linawolf (approved):

linawolf:
@jaapio please rebase and resolve the conflicts

PR #24, opened 2023-11-02
[TAKS] Update phpdocumentor/guides

linawolf:
the Markdowns support is better now and the :options: directive available


linawolf:
https://github.com/TYPO3-Documentation/render-guides/pull/25 should fix the integration tests

brotkrueml (approved):

PR #29, opened 2023-11-02
[FEATURE] Implement ":issues:" directive

linawolf:
This directive is used in the changelog and other places to link to forge.

### DIFF
--- a/packages/typo3-docs-theme/src/TextRoles/IssueReferenceTextRole.php
+++ b/packages/typo3-docs-theme/src/TextRoles/IssueReferenceTextRole.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace T3Docs\Typo3DocsTheme\TextRoles;
+
+use phpDocumentor\Guides\Nodes\Inline\AbstractLinkInlineNode;
+use phpDocumentor\Guides\Nodes\Inline\HyperLinkNode;
+use phpDocumentor\Guides\RestructuredText\TextRoles\AbstractReferenceTextRole;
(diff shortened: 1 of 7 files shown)


packages/typo3-docs-theme/src/TextRoles/IssueReferenceTextRole.php
```diff
@@ -0,0 +1,38 @@
...
+    /** @inheritDoc */
+    public function getAliases(): array
+    {
+        return [];
+    }
+
+    /** @return HyperLinkNode */
+    protected function createNode(string $referenceTarget, string|null $referenceName, string $role): AbstractLinkInlineNode
+    {
+        return new HyperLinkNode($referenceName ?? sprintf('forge#%s', (int)$referenceTarget), sprintf('https://forge.typo3.org/issues/%s', (int)$referenceTarget));
```
brotkrueml, line 36: I would prefer to define the "magic" string (https://forge.typo3.org/...) as a class constant. This way, it is easy to spot instead of scrolling and reading code. What do you think?

Maybe also the "forge%s" thingy?

Additionally, when using sprintf, you can use %d (treated as integer) instead of %s, so you don't have to cast to an int.

brotkrueml (changes requested):

packages/typo3-docs-theme/src/TextRoles/IssueReferenceTextRole.php
linawolf: done, thanks for the suggestions

brotkrueml (approved):

PR #30, opened 2023-11-02
[FEATURE] Add version switch to template

jaapio:
Versions are loaded via javascript, we just need to render the current version.


jaapio:
We will need a change in guides to set the project-version from the commandline to make it easier to set this via a github action.

linawolf (approved):

PR #31, opened 2023-11-02
[FEATURE] add configure command adjust config

jaapio:


brotkrueml (commented):
I would prefer to add a description to the command which explains what it does. What do you think?

jaapio:
Yes, that should be added, I have some more plans with this command. Like setting up the default config and directory structure.

garvinhicking:
I continued this here and added it into our toolbox:

https://github.com/TYPO3-Documentation/render-guides/commit/9d07b3219817da9b0859b1849e3da982c6f4b87a

garvinhicking:
Closing with the recent changes this is adressed.

PR #38, opened 2023-11-03
[TASK] Move fixtures into integration tests

linawolf:
Now that we have integration tests available these can be automatically tested. The directory "fixtures" is not needed anymore


tests/Integration/tests/markdown/expected/README.html
```diff
@@ -0,0 +1,279 @@
...
+
+<h2>Installation</h2>
+
+<p><img src="https://upload.wikimedia.org/wikipedia/commons/a/a3/Forgiveness_5.gif" alt="Swing Image"/></p>
+
+    
+<ol>
+    <li><p><strong>Choose a Sturdy Location:</strong> Select a location with a sturdy beam or swingset structure capable of supporting the weight of the swing and the child.</p></li>
+
+    <li><p><strong>Securely Attach the Swing:</strong> Use strong and reliable <a href="https://example.com/ropes">ropes</a>or chains to securely attach the swing to the chosen location. Double-checkthe knots or hardware to ensure they are tight and safe.</p></li>
```
FlorianT03, line 128: A space must be inserted after the word "ropes", as the output of the code displays "ropesor" at this point, so a small typo

linawolf: This seems to be standard behaviour of Markdown and can be fixed by putting a &nbsp; here or two empty spaces at the end of the line (which do however get removed by IDEs unfortunary)

linawolf: I solved it like that now in the markdown. good catch

FlorianT03: I see. Good to know^^

FlorianT03 (approved):
After my one comment I saw no other mistake, so everything is fine in my eyes!

PR #43, opened 2023-11-05
[FEATURE] Introduce php:interface textrole

linawolf:


packages/php-domain/src/TextRoles/InterfaceTextRole.php
```diff
@@ -0,0 +1,109 @@
...
+    public function getName(): string
+    {
+        return self::NAME;
+    }
+    public function getAliases(): array
+    {
+        return [];
+    }
+
+    public function processNode(
```
jaapio, line 36: should we maybe extract this to a trait in guides to make is reusable? This seems to me a very complex piece of code, and we are reusing this very often.

linawolf: We do have it in a the GeneralLinkTextRole, however we have a special case as all other textroles are escaping the links (\\ -> \) while this text role does not want to escape backslashes. So we cannot reuse the code.

jaapio (approved):

PR #44, opened 2023-11-05
[TASK] Sanitize input for youtube directive

linawolf:


garvinhicking (approved):
Looking good to me! As we talked about this, maybe adding another test case to showcase the 11-character limit would be nice, maybe that's not really necessary.

PR #45, opened 2023-11-05
[TASK] Add tests for bootstrap cards

linawolf:
We currently do it like this: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Content/Cards.html Though in future I would like to introduce a directive for that.


FlorianT03 (approved):
Looks fine to me

PR #48, opened 2023-11-06
[DOCS] Fix docker command in readme

brotkrueml:
Currently, running the command fails with the error:

    Too many arguments, expected arguments "input".

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,7 +10,7 @@ Usage with Docker
 
 ::
     mkdir Documentation-GENERATED-temp
-    docker run --rm --pull always -v $(pwd):/project -it ghcr.io/typo3-documentation/render-guides:main Documentation Documentation-GENERATED-temp
+    docker run --rm --pull always -v $(pwd):/project -it ghcr.io/typo3-documentation/render-guides:main --output Documentation-GENERATED-temp Documentation
 
 Contributing
 ============


README.rst
```diff
@@ -10,7 +10,7 @@ Usage with Docker
 
 ::
     mkdir Documentation-GENERATED-temp
-    docker run --rm --pull always -v $(pwd):/project -it ghcr.io/typo3-documentation/render-guides:main Documentation Documentation-GENERATED-temp
+    docker run --rm --pull always -v $(pwd):/project -it ghcr.io/typo3-documentation/render-guides:main --output Documentation-GENERATED-temp Documentation
```
linawolf, line 13: is there a specific reason why you move the option to the front and the argument to the back? Isn't the default other way round? also it should be --output="Documentation-GENERATED-temp" I think?

jaapio: The order of options does not matter in symfony console. but it is common to use options before arguments.

brotkrueml: I choose to being close to the note given when running in the error:

    run [--output OUTPUT] [--input-file INPUT-FILE] [--input-format INPUT-FORMAT] [--output-format OUTPUT-FORMAT] [--log-path LOG-PATH] [--fail-on-log] [--theme THEME] [--progress|--no-progress] [--] [<input>]

The input is at the end.

linawolf (approved):

PR #50, opened 2023-11-08
[TASK] Update guides and Improve template registration

linawolf:
@jaapio please have a look why this is not working


packages/php-domain/src/DependencyInjection/PhpDomainExtension.php
```diff
@@ -5,30 +5,41 @@
...
+use T3Docs\PhpDomain\Nodes\FullyQualifiedNameNode;
+use T3Docs\PhpDomain\Nodes\PhpComponentNode;
+use T3Docs\PhpDomain\Nodes\PhpNamespaceNode;
+
 final class PhpDomainExtension extends Extension implements PrependExtensionInterface
 {
     /** @param mixed[] $configs */
     public function load(array $configs, ContainerBuilder $container): void
     {
+        $container->prependExtensionConfig(
```
jaapio, line 26: You need to put these lines in 'prepend(ContainerBuilder $container): void {}' otherwise it will not work. 

The load method cannot extend the configuration.

jaapio (approved):

PR #52, opened 2023-11-08
[FEATURE] Add helper script to inject git hooks (commit-msg+pre-commit)

garvinhicking:
This allows contributors to execute:

`make githooks` (or: `./tools/add-githooks.sh` directly)

This in turn copies the tools/githook-templates/ files to the local git directory for the project.

That in turn executes commit message line length inspection and executes `make pre-commit-test` before any commit.

TODO: At least for me sadly currently `make pre-commit-test` (macOS) does not work, complains about a missing entry point. But this is independent from what this commit does.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,10 @@ phpstan-baseline:
 phpstan:
 	$(PHP_BIN) -d memory_limit=1024M vendor/bin/phpstan --configuration=phpstan.neon
 
+.PHONE: githooks
+githooks: ## Runs script that injects githook pre-commit, so that 'make pre-commit-test' is performed on each commit.
+	./tools/add-githooks.sh
+
 .PHONY: test
 test: test-integration test-unit test-docs## Runs all test suites with phpunit/phpunit
 
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-PHP_BIN = docker run -it --rm --user $$(id -u):$$(id -g) -v${PWD}:/opt/project -w /opt/project php:8.2-cli php -d memory_limit=1024M
+PHP_BIN ?= docker run -it --rm --user $$(id -u):$$(id -g) -v${PWD}:/opt/project -w /opt/project php:8.2-cli php -d memory_limit=1024M
+## This can be adapted in the .git/hooks/pre-commit call of this step. If you change this line (i.e. new PHP version), also change it in that template
 
 .PHONY: help
 help: ## Displays this list of targets with descriptions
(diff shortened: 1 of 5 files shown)


linawolf:
Unfortunary for me it doesnt work either. On commiting then I get the error:

Checking coding guidelines via 'make pre-commit-test' ...
docker run -it --rm --user $(id -u):$(id -g) -v/home/lina/projects/t3/rendering/render-guides:/opt/project -w /opt/project php:8.2-cli php -d memory_limit=1024M vendor/bin/php-cs-fixer fix
the input device is not a TTY
make: *** [Makefile:13: fix-code-style] Error 1

garvinhicking:
Sorry, I somehow didn't see this comment from you. We have an annoying problem now.


The problem with this is that the Makefile currently does this:

````
PHP_BIN = docker run -it --rm --user $$(id -u):$$(id -g) -v${PWD}:/opt/project -w /opt/project php:8.2-cli php -d memory_limit=1024M
````

The option "-t" (within "-it") wants to allocate a TTY; if you run the make command interactively, that's ok. If it's run by the git hook though, that runs without an interactive tty, and there, the "-t" is an issue.

Removing the "-t" option from the general configuration would work, but then running the script manually will no longer have colors in the output (and I don't know if any of the other tasks in the Makefile actually require an interactive shell?).

So my solutions would be:

a.) Provide a "-no-tty" variant of EVERY task that currently uses PHP_BIN and that we need for the "pre-commit-tests", and make those not use the "-t" option. That would mean adding fix-code-style-no-tty, test-no-tty, code-style-no-tty, static-code-analysis-no-tty and test-mono-repy-no-tty. That's kinda bloaty, but I could do this.

b.) Somehow(tm) find a way to let the "make pre-commit-test" command have a second option like "make pre-commit-test --no-tty" and pass on that option to the Makefile, and then conditionally either use `-i` or `-it` for the PHP_BIN constant. Sadly my Makefile skills are not sufficient for this.

c.) Completely ignore the Makefile and add the proper commands into the git commit hook file. That would also in the future allow us to do a bit shorter git pre-commit checks since they no longer need to be in sync with the Makefile. But that's also the drawback - whenever we adapt the Makefile, we need to remember the githook, too.

Thoughts?

jaapio:
If you change the assignment of PHP_BIN to `PHP_BIN ?=` 
you can pass a new value like this:
`make PHP_BIN=php test`

garvinhicking:
@jaapio Awesome, many thanks for letting me know! I've updated the commit, now it seems to work locally.

Not totally a fan of redundance of having to maintain PHP_BIN at two places then, but I can't think of a more convenient way atm...

linawolf (approved):

linawolf:
This works now for me also. Great work, thank you!

PR #54, opened 2023-11-08
[TASK] Add integration tests for existing documentation options

linawolf:


tests/Integration/tests/configuration/input/guides.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<guides xmlns="https://www.phpdoc.org/guides"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
```
garvinhicking, line 4: I think this should be a valid, callable URI, no? Because it isn't, even when removing the space in the URI?

tests/Integration/tests/configuration/expected/objects.inv.json
```diff
@@ -0,0 +1,18 @@
+{
+    "std:doc": {
```
garvinhicking, line 2: Could you explain what the difference is of std:doc and std:label, because the contain the same values?

tests/Integration/tests/configuration/input/guides.xml
```diff
@@ -0,0 +1,13 @@
...
+        links-are-relative="true"
+        default-code-language="php"
+>
+    <project title="My Project" version="3.1.4" />
+    <inventories>
+        <inventory id="h2document" url="https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/" />
+        <inventory id="t3coreapi" url="https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/" />
```
garvinhicking, line 11: According to https://github.com/phpDocumentor/guides/blob/main/packages/guides-cli/resources/schema/guides.xsd the `<inventory>` tag would need to have a `path` attribute, not `url`? And not `id` but `name`?

tests/Integration/tests/configuration/input/guides.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<guides xmlns="https://www.phpdoc.org/guides"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
+        links-are-relative="true"
```
garvinhicking, line 5: According to the XSD that attribute doesn't exist?

Also, these are listed but not set in the example:

* input
* input-file
* output
* input-format
* log-path
* fail-on-log
* show-progress
* theme

Do we want to expose these?

tests/Integration/tests/configuration/input/guides.xml
```diff
@@ -0,0 +1,13 @@
...
+        default-code-language="php"
+>
+    <project title="My Project" version="3.1.4" />
```
garvinhicking, line 8: The XSD lists `<extension>` however that one seems a bit dangerous when exposing to be interpreted by foreign extensions, what do you think?

Also a `<theme extends="">` tag exists (don't understand how it relates to the `<guides theme="">` attribute)?

Last a `<template>` tag is mentioned in the XSD (with attributes file, node, format) - is that something we want to expose also to documentation modules?

garvinhicking (commented):
Thanks for this one, that helps to talk about the structure. :-)

tests/Integration/tests/configuration/expected/objects.inv.json
linawolf: In a rst project consisting of several documents each document is listed in std:doc it can then be used to link to a document using its path. i.e. ``:doc:`index` ``

std:label is linking to each headline (we only have one in this example). If there is more then one headline in one Document=rst file you can link to each headline. you can also link to elements that have an explicit anchor defined.

The file is used when we are linking between different manuals / projects.

It also contains the version and manual/project title that was set in the guides.xml, that is why I included it in the tests

tests/Integration/tests/configuration/input/guides.xml
linawolf: I must check that, there might be a problem in the xsd file then.  links-are-relative should exist.

In our usecase it does not make much sense to set the other values. We are either setting them in the command (--output-path=xy) or in the global guides.xml or the default is good to go.

tests/Integration/tests/configuration/input/guides.xml
linawolf: Only extensions that are already loaded via composer can be included with the <extension> tag. So while i would not suggest anyone to tinker with that it is impossible to include an extension that is not already loaded. You could theoretically exclude extensions however -  like if you don't need plant-uml you could exclude that extension to be faster or whatever.

We could, of course also define our own configuration format just for the theme later on

tests/Integration/tests/configuration/input/guides.xml
linawolf: Mmh, we are doing it like this in the tests, and it only works like that. https://github.com/phpDocumentor/guides/blob/main/tests/Integration/tests/navigation/intersphinx-link/input/guides.xml

I think we need to update the scheme in the guides and add some automatic testing against the scheme in our testing workflow in the guides...

tests/Integration/tests/configuration/expected/objects.inv.json
fe-hicking: Ah, I see. Those would not be part of the settings.cfg migration, or are they?!

tests/Integration/tests/configuration/input/guides.xml
fe-hicking: Oh, I didn't know they diverged. Looks like a breaking change, too, so a new xsd revision would need to be created? Do you know how phpDocumentor would handle this?

linawolf: We did not yet publish v1.0 of the guides so we can still make breaking changes without a problem.

tests/Integration/tests/configuration/input/guides.xml
garvinhicking: Never mind, I misread the docs. It's right.

tests/Integration/tests/configuration/input/guides.xml
garvinhicking: (Stalled for now, because we can't get the XSD validation to work properly on the existing XML+XSD pairs of the guides repo)

tests/Integration/tests/configuration/expected/objects.inv.json
linawolf: No, they are output and not input

tests/Integration/tests/configuration/input/guides.xml
linawolf: I believe, this is fixed by https://github.com/phpDocumentor/guides/pull/668

tests/Integration/tests/configuration/input/guides.xml
linawolf: The open issues should be resolved by https://github.com/phpDocumentor/guides/pull/666

Makefile
```diff
@@ -44,6 +44,11 @@ test-integration: ## Runs integration tests with phpunit/phpunit
 test-docs: ## Generate projects docs without warnings
 	$(PHP_BIN) vendor/bin/guides -vvv --no-progress Documentation --output="/tmp/test" --fail-on-log
 
+.PHONY: test-xml
+test-xml: ## Lint all guides.xml
+	./tools/xmllint.sh
```
jaapio, line 49: Please add this to the QA workflow on github, so this is validated before we can merge.

jaapio (changes requested):

Makefile
linawolf: Can we do that? It is pure linux?

jaapio: Yes, github actions are executed on a full Ubuntu stack.

linawolf: would I just call the script then somehow in the github actions?

jaapio: You can call this script in the workflow like this:

```
      - name: "Lint xml configs"
        run: "tools/xmllint.sh"
```

linawolf: added it, please have a look

linawolf:
@jaapio xmlint is not found in the github action, what can we do about that?

garvinhicking:
@linawolf Seems to work with "sudo" and the xmllint integration.

Note to self: If we merge the Makefile-PR we'll need to also run the xmllint inclusion there (easy)

garvinhicking (approved):
I believe this works; the change in guides.xml with a different syntax for  the extension specification isn't easily reviewable for me though, but documentation generation still works for me, it seems.

Before:
````
<phpdomain>
  <class>\T3Docs\PhpDomain\DependencyInjection\PhpDomainExtension</class>
</phpdomain>
````

After:
```
<extension class="\T3Docs\PhpDomain\DependencyInjection\PhpDomainExtension"/>
```

jaapio (approved):

PR #56, opened 2023-11-09
[TASK] relocate config to Documentation

jaapio:
The default location for guides.xml is Documentation, people might decide to do it different, so make it configurable.

I update the entrypoint to fix an issue when the input argument was omitted. The previous implementation did quote all arguments. But that broke the optional input argument as it would always be `""` when left empty. 

While testing this behavior I found another issue, that was giving some odd errors when a user does not specify a volume, we to now exit with an error message.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -4,7 +4,13 @@ description: 'Documentation render tool for typo3 documentation'
 inputs:
   input:
     description: 'Input directory'
+    required: false
+    default: null
+
+  config:
+    description: 'Configuration file location'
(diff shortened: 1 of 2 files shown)


linawolf:
@jaapio, if I understand that correctly you have to set where is the configuration file, in the docker command now? That would mean, we need to update the README of this project?

jaapio:
Good question, I think we should do that, it's only required when the project has a documentation folder.

I will update this in the open documentation pr.

PR #61, opened 2023-11-10
[TASK] Add guides.xml in the documentation dir

jaapio:
Guides.xml at root level is used as default for all projects, our specific config is in Documentation.

### DIFF
--- a/Documentation/guides.xml
+++ b/Documentation/guides.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<guides
+    xmlns="https://www.phpdoc.org/guides"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="https://www.phpdoc.org/guides ../vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
+>
+    <project title="Render guides" />
+</guides>
--- a/guides.xml
+++ b/guides.xml
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<guides links-are-relative="true">
+<guides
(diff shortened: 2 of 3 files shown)


guides.xml
```diff
@@ -1,5 +1,9 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<guides links-are-relative="true">
+<guides
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
linawolf, line 3:
```suggestion
    xmlns="https://www.phpdoc.org/guides"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```

We need to put the namespace or it will not validate

Documentation/guides.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<guides
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
linawolf, line 3:
```suggestion
    xmlns="https://www.phpdoc.org/guides"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
We need to put the namespace or it will not validate

Documentation/guides.xml
```diff
@@ -0,0 +1,8 @@
...
+    xsi:schemaLocation="https://www.phpdoc.org/guides ../vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
+    links-are-relative="true"
```
linawolf, line 5:
```suggestion
    links-are-relative="true"
```

I think this should be always the default in the guides and not be set by the users

linawolf (commented):
Just some small Remarks

linawolf (approved):

PR #66, opened 2023-11-10
[FEATURE] Render Methods in interfaces

linawolf:


garvinhicking (commented):
This looks very solid to me, but I could only dry-read it and not verify. Can do that next week if it's not approved by then 👍

tests/Integration/tests/php/interface-with-method/input/index.rst
```diff
@@ -0,0 +1,18 @@
+=====================================
+PHP Interface with implicit namespace
+=====================================
+
+..  php:interface:: TYPO3\CMS\Core\TestInterface
+
+    Lorem Ipsum Dolor!
+
+    ..  php:method:: setDate(int $year, int $month, int $day)
```
brotkrueml, line 9: Is it intended that a trailing comma is rendered after `$day`?

image

packages/php-domain/src/PhpDomain/MethodNameService.php
```diff
@@ -0,0 +1,27 @@
...
+
+declare(strict_types=1);
+
+namespace T3Docs\PhpDomain\PhpDomain;
+
+use T3Docs\PhpDomain\Nodes\MethodNameNode;
+
+class MethodNameService
+{
+    private const PATTERN = '/^\s*(\w+)\s*\(\s*(.*?)\s*\)\s*(?::\s*(\w+))?\s*$/';
```
brotkrueml, line 11: Please provide a regex101.com link and provide there some examples. This makes it easier to get what the regex does.

There is also a phpstan rule which enforces this in the package symplify/php-stanrules:
https://github.com/symplify/phpstan-rules/blob/main/docs/rules_overview.md#annotateregexclassconstwithregexlinkrule
Maybe we should add this.

packages/php-domain/src/PhpDomain/MethodNameService.php
```diff
@@ -0,0 +1,27 @@
...
+
+    public function getMethodName(string $name): MethodNameNode
+    {
+        if (preg_match(self::PATTERN, $name, $matches)) {
```
brotkrueml, line 15: We should log a warning if the pattern does not match.

brotkrueml (changes requested):

packages/php-domain/src/PhpDomain/MethodNameService.php
linawolf: Yes, why not. Would you like to make a PR for that rule?

packages/php-domain/src/PhpDomain/MethodNameService.php
linawolf: done

packages/php-domain/src/PhpDomain/MethodNameService.php
linawolf: I added the comment

tests/Integration/tests/php/interface-with-method/input/index.rst
linawolf: Good catch! changed it

packages/php-domain/src/PhpDomain/MethodNameService.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace T3Docs\PhpDomain\PhpDomain;
+
+use Psr\Log\LoggerInterface;
+use T3Docs\PhpDomain\Nodes\MethodNameNode;
+
+class MethodNameService
+{
+    /** https://regex101.com/r/QrrSXk/1 */
+    private const PATTERN = '/^\s*(\w+)\s*\(\s*(.*?)\s*\)\s*(?::\s*(\w+))?\s*$/';
```
brotkrueml, line 13: To be valid with #68 the constant has to be suffed with `_REGEX`. I would also suggest to use another name than `PATTERN`, maybe `METHOD_SIGNATURE_REGEX`?

packages/php-domain/src/PhpDomain/MethodNameService.php
```diff
@@ -0,0 +1,34 @@
...
+            $methodName = $matches[1];
+            $parameters = isset($matches[2]) ? $matches[2] : '';
+            $returnType = $matches[3] ?? null;
+            $parametersArray = preg_split('/\s*,\s*/', $parameters, -1, PREG_SPLIT_NO_EMPTY);
+            if ($parametersArray === false) {
+                $parametersArray = [];
+            }
+            return new MethodNameNode($methodName, $parametersArray, $returnType);
+        }
+        $this->logger->warning(sprintf('Method name %s is invalid. See https://regex101.com/r/QrrSXk/1 for which kind of names are allowed.', $name));
```
brotkrueml, line 31:
```suggestion
        $this->logger->warning(sprintf('Method signature %s is invalid. See https://regex101.com/r/QrrSXk/1 for which kind of names are allowed.', $name));
```

As far as I understand the signature consists of the method name, the arguments and the return type. So this would be more appropriate.

Additionally, I suggest to remove the regex101 link in the warning: We allow all method signatures PHP processes. Also the link can change, when the regex examples change (mind the `/1` at the end of the URL which marks the version).

brotkrueml (changes requested):

packages/php-domain/src/PhpDomain/MethodNameService.php
linawolf: done

packages/php-domain/src/PhpDomain/MethodNameService.php
```diff
@@ -0,0 +1,34 @@
...
+            $methodName = $matches[1];
+            $parameters = isset($matches[2]) ? $matches[2] : '';
+            $returnType = $matches[3] ?? null;
+            $parametersArray = preg_split('/\s*,\s*/', $parameters, -1, PREG_SPLIT_NO_EMPTY);
+            if ($parametersArray === false) {
+                $parametersArray = [];
+            }
+            return new MethodNameNode($methodName, $parametersArray, $returnType);
+        }
+        $this->logger->warning(sprintf('Method siganture %s is invalid. ', $name));
```
brotkrueml, line 31:
```suggestion
        $this->logger->warning(sprintf('Method signature %s is invalid. ', $name));
```

brotkrueml (approved):

PR #68, opened 2023-11-12
[TASK] Enforce regex validation link

brotkrueml:
With the help of the phpstan rule "AnnotateRegexClassConstWithRegexLinkRule" a link is necessary for easy access what a regex does and which strings are validated or not. This makes is easier to grasp a regex syntax and to avoid errors.

Regexes have to be a constant with the suffix _REGEX.

### DIFF
--- a/packages/php-domain/src/TextRoles/InterfaceTextRole.php
+++ b/packages/php-domain/src/TextRoles/InterfaceTextRole.php
@@ -14,7 +14,13 @@ use Psr\Log\LoggerInterface;
 
 final class InterfaceTextRole implements TextRole
 {
+    /**
+     * @see https://regex101.com/r/OyN05v/1
+     */
+    private const INTERFACE_NAME_REGEX = '/^([a-zA-Z0-9]+):(.*$)/';
+
     private readonly InlineLexer $lexer;
+
     public function __construct(
         private readonly LoggerInterface $logger,
         private readonly AnchorReducer $anchorReducer,
@@ -95,8 +101,7 @@ final class InterfaceTextRole implements TextRole
     /** @return ReferenceNode */
     protected function createNode(string $referenceTarget, string|null $referenceName, string $role): AbstractLinkInlineNode
     {
-        $pattern = '/^([a-zA-Z0-9]+):(.*$)/';
-        if (preg_match($pattern, $referenceTarget, $matches)) {
+        if (preg_match(self::INTERFACE_NAME_REGEX, $referenceTarget, $matches)) {
             $interlinkDomain = $matches[1];
             $id = $this->anchorReducer->reduceAnchor($matches[2]);
         } else {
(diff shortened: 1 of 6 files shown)


brotkrueml:
@linawolf 

Can you have a look at FullyQualifiedNameService::FULL_NAME_PATTERN_REGEX? I do not think that the regex works as expected (see the regex101.com examples).

Additionally, I don't really understand for what cases the InterfaceTextRole::INTERFACE_NAME_REGEX is needed. Aas far as I understand, "php:interface" is matched - and only that should match? Or am I wrong?

packages/php-domain/src/TextRoles/InterfaceTextRole.php
```diff
@@ -14,7 +14,13 @@
 
 final class InterfaceTextRole implements TextRole
 {
+    /**
+     * @see https://regex101.com/r/OyN05v/1
+     */
+    private const INTERFACE_NAME_REGEX = '/^([a-zA-Z0-9]+):(.*$)/';
```
linawolf, line 20:
```suggestion
    private const INTERLINK_NAME_REGEX = '/^([a-zA-Z0-9]+):(.*$)/';
```

We are matching for interlinks here which would be devided by a colon from the interface name. We are not matching for interfaces. This can be used to link to an interface in another manuals i.e. 

```
:php:interface:`someManual:\MyVendor\MyProject\SomeInterface`

**Comment by linawolf** [comment]
> Can you have a look at FullyQualifiedNameService::FULL_NAME_PATTERN_REGEX? I do not think that the regex works as expected (see the regex101.com examples).

To me it seems like you did not remove the escaping of the backslash that is due to the PHP string? What is `'\\'` in PHP would only be  `\` on regex101, therefore 4 backslashes then become only two and mean the backslash character?

**Comment by brotkrueml** [comment]
> > Can you have a look at FullyQualifiedNameService::FULL_NAME_PATTERN_REGEX? I do not think that the regex works as expected (see the regex101.com examples).
> 
> To me it seems like you did not remove the escaping of the backslash that is due to the PHP string? What is `'\\'` in PHP would only be `\` on regex101, therefore 4 backslashes then become only two and mean the backslash character?

So both cases are valid (with and without the starting `\`)?

    MyVendor\\MyLibrary\\MyClass
    \\MyVendor\\MyLibrary\\MyClass

Don't care in this case if one or two backslashes, it's here only about regex validation.

File: packages/php-domain/src/TextRoles/InterfaceTextRole.php [hunk 2]
```diff
@@ -95,8 +101,7 @@ public function processNode(
     /** @return ReferenceNode */
     protected function createNode(string $referenceTarget, string|null $referenceName, string $role): AbstractLinkInlineNode
     {
-        $pattern = '/^([a-zA-Z0-9]+):(.*$)/';
-        if (preg_match($pattern, $referenceTarget, $matches)) {
+        if (preg_match(self::INTERFACE_NAME_REGEX, $referenceTarget, $matches)) {
```
brotkrueml, line 104:
```suggestion
        if (preg_match(self::INTERLINK_NAME_REGEX, $referenceTarget, $matches)) {
```

linawolf (approved):

PR #69, opened 2023-11-13
[BUGFIX] Fix failing regex validation

brotkrueml:
This fixes the phpstan error:

    ------ -----------------------------------------------------------------------
     Line   php-domain/src/PhpDomain/MethodNameService.php
    ------ -----------------------------------------------------------------------
     13     Add regex101.com link to that shows the regex in practise, so it will
            be easier to maintain in case of bug/extension in the future
    ------ -----------------------------------------------------------------------

Related: #66 
Related: #68


linawolf (approved):

linawolf:
Thanks for catching this, I should have rebased before merging

Issue #72, opened 2023-11-14
Executing docker on typo3-docs:local does not evaluate bash arguments?


garvinhicking:
I created my custom typo3-docs:local docker container via:

`docker build -t typo3-docs:local .`

That works. However when I call this:

`docker run --rm --user=$(id -u):$(id -g) --volume ${PWD}:/project typo3-docs:local --version` 

I would expect the output showing the version of the called `entrypoint.sh` marker, in the end running this inside the container:

`/opt/guides/vendor/bin/guides --version`

That works if I execute it in my manual shell. But even when I add the output to the `entrypoint.sh` file, I can see that `--version` is passed along, but still it doesn't seem to take effect:

````bash
  # ... snippet of entrypoint.sh at the end
  echo "Inline bash execution with commands $@"
  echo sh -c "$@"
  sh -c "$@"
````

This yields:

```
Inline bash execution with commands /opt/guides/vendor/bin/guides --version
sh -c /opt/guides/vendor/bin/guides --version
  0/10 [>---------------------------]   0%

In menu_doc.html.twig line 6:

  An exception has been thrown during the rendering of a template ("No root document entry was found").


In ProjectNode.php line 133:

  No root document entry was found


run [--output OUTPUT] [--input-file INPUT-FILE] [--input-format INPUT-FORMAT] [--output-format OUTPUT-FORMAT] [--log-path LOG-PATH] [--fail-on-log] [--theme THEME] [--progress|--no-progress] [--] [<input>]
```

(Don't mind the error message, that is due to Issue #53)

**Comment by garvinhicking** [comment]
I do not know the specific reason why this is how it is, but it's fixed if I changed the entrypoint.sh file to do this:

```
php "$@"
```

instead of this:

```
sh -c "$@"
```

This is adressed in PR #73

garvinhicking:
(Resolved through PR 73, at least on macOS :) )

PR #73, opened 2023-11-14
[TASK] [!!!] Utilize Makefile as command runner

garvinhicking:
Makefile is now the central "source of truth". It contains all commands (targets), possible compound commands, as well as an optional environment option. By default those are executed in a docker instance, i.e.::

```
make docs
```

By utilizing an environment option (ENV=local or ENV=docker [default]) the selected Makefile command can be either executed locally, within DDEV or witin the prebuild/custom docker container. 

So to prevent using Docker you can run:

```
make docs ENV=local
```

All make targets can be routed through composer scripts; this routing is what is allowing ddev to easily execute scripts:

```
ddev composer make docs
```

Composer automatically passes along the "ENV=local" option for any wildcard task, thus composer does not need to know of any specific task at all.

The ddev config now requires the "build-essential" package so that "make" is available there.

The github actions were changed from composer scripts to Makefile commands.

This needs proper review and testing before (or: if) getting merged, it has the capacity to break all processes.

I would suggest to perform these changes rather sooner than later, to prevent future problems. Currently the build/dev/contribute pipeline is a mixture of Makefile / composer scripts and docker commands and are hard to distinguish.

P.S.: I also lowered the PHP version to "8.1" to comply with our lowest supported version. Of course that can only be enforced via ddev+docker, not on "local" executions.


linawolf:
@jaapio it would be great if you could have a look at this

linawolf:
@garvinhicking when I run `make pre-commit-test` locally I get the following error:
```
ENVIRONMENT: Docker (default)
docker run -i --rm --user $(id -u):$(id -g) -v/home/lina/projects/t3/rendering/render-guides:/opt/project -w /opt/project php:8.1-cli php -d memory_limit=1024M composer validate --no-check-publish
Could not open input file: composer
```

garvinhicking:
> @garvinhicking when I run `make pre-commit-test` locally I get the following error:
> 
> ```
> ENVIRONMENT: Docker (default)
> docker run -i --rm --user $(id -u):$(id -g) -v/home/lina/projects/t3/rendering/render-guides:/opt/project -w /opt/project php:8.1-cli php -d memory_limit=1024M composer validate --no-check-publish

OK, I messed up at this point: PHP_BIN points to a docker php image. That one does not have composer.

PHP_PROJECT_BIN points to "our" project container. That one has PHP and composer. But its entrypoint exclusively allows to run `bin/guides`, so it can NOT be used to run composer or any other native PHP command.

Thus I've now changed the two tasks that run composer validate and install to use the `composer:latest` docker image. Probably we should fix that to a specific version, if this patch gets merged. This allows to execute the composer validation by mounting the project on /app. And if the command is executed locally via `make pre-commit-test ENV=local` it should just use the local composer binary.

Makefile
```diff
@@ -1,74 +1,142 @@
...
+.PHONY: show-env
+show-env: ## Shows PHP environment options (buildinfo)
+	@echo "$(ENV_INFO)"
+	@echo "Base PHP:"
+	$(PHP_BIN) --version
+	@echo ""
+
+	@echo "Project within Docker:"
+    # TODO: See Issue #72, does not work yet
+	docker run --rm --user $$(id -u):$$(id -g) -v${PWD}:/project typo3-docs:local --version
```
jaapio, line 49: `--user` must be omitted, as the docker image will detect the correct user.

README.rst
```diff
@@ -3,60 +3,69 @@
...
-    make pre-commit-test
+::
 
-You can use a helper script to set this up once in your project::
+    # Build the custom local Docker container
+    docker build --file Dockerfile --tag typo3-docs:local .
 
-    make githooks
+    # Execute the Docker container that is provided locally, build Documentation
+    docker run --rm --user=$(id -u):$(id -g) --volume ${PWD}:/project typo3-docs:local ./Documentation ./Documentation-GENERATED-temp --theme=typo3docs
```
jaapio, line 38: please omit `--user`,  the output directory is changed to `--output` and can be omitted, as we set this by default, same applies for the input director, which is automatically detected. 

We can simplify this to:

`docker run --rm --user=$(id -u):$(id -g) --volume ${PWD}:/project --progress`

README.rst
jaapio, line 1: Should this file not be replaced with the default readme format like other documentation projects have?

jaapio (commented):
I do not see any blocking issues in this pr, some small adjustments that could make it easier to use, and corrections to make commands in the documentation work.

Interesting to be aware of, the Makefile has a help target to make the user aware of possible options. And we are also writing documentation, from my point of view this raises the maintenance effort. My focus with phpDocumentor is to try to automate this stuff away. I like inline documentation being extracted in a certain way. I'm not sure yet what this could mean for this particular case, but I would like to drop this seed in to your minds as well. 
This does not mean that I disagree with the duplication and more extended way of documenting the contribution guidelines, but I would like to think about this. 

I like the way you centralized the make file as a single source! Thank you for opening my eyes for this!

Makefile
garvinhicking: Hm, if I change that to:

`docker run --rm -v${PWD}:/project typo3-docs:local --version`

it only reports:

```
ould not detect the owner of /project. Did you mount your project?.
Run this container with "docker run --rm  --volume ${PWD}:/project ghcr.io/typo3-documentation/render-guides:main"
```

(note that I ran the above with "typo3-docs:local" - and did you check out Issue #72 - this has some more background?)

README.rst
garvinhicking: You probably mean:

`docker run --rm --user=$(id -u):$(id -g) --volume ${PWD}:/project typo3-docs:local --progress`

(note the "typo3-docs:local" in there). However, when I run this with "--version" instead of "--progress" I still do not get the output of the interpreted --version command?!

garvinhicking:
Hi @jaapio 

> I do not see any blocking issues in this pr, some small adjustments that could make it easier to use, and corrections to make commands in the documentation work.

Phew, thanks for that initial raise of spirits :)

> Interesting to be aware of, the Makefile has a help target to make the user aware of possible options. And we are also writing documentation, from my point of view this raises the maintenance effort. My focus with phpDocumentor is to try to automate this stuff away. I like inline documentation being extracted in a certain way. I'm not sure yet what this could mean for this particular case, but I would like to drop this seed in to your minds as well. This does not mean that I disagree with the duplication and more extended way of documenting the contribution guidelines, but I would like to think about this.

That's a very good "nudge". Yes, indeed there is some redundancy/verbosity. The "make help" command only lists a few makefile target descriptions, I figured that to be "best practice". Not sure how we could have only one source of these commands, but I'll reflect on it, thank you!

README.rst
garvinhicking: I'm just a "new guy" with little experience, that's a question for Lina and Chris 🥶

linawolf:
So the make pre-commit command ist now working for me locally. I cannot help with the other issues unfortunary

garvinhicking:
To recap the state of this PR:

* The issue with macOS filenames has been addressed, and needs a PR merge of https://github.com/phpDocumentor/guides/pull/677 in the `guides` project
* The issue with failing to pass arguments (Issue #72) is solved with this PR
* Jaapios suggestion to omit "--user" arguments doesn't work for me, but it works when using it, so I left it at that currently
* The command "make show-env" now works
* The command "make pre-commit-test" now works in both local and docker env
* Shortened shortcut commands listed in README.rst to use default arguments

Unresolved:

* Suggested by Jaapio: Do we want to keep README.rst, or should the repository have a README.md file instead?

linawolf:
I would keep the README in rst as we in the documentation team generally favor rst

README.rst
linawolf: We usually have a README.rst as wo want ppl to use rst

jaapio (approved):

garvinhicking:
(@brotkrueml maybe we want to discuss this in person next week, I'm open to refactoring this to suit the project's needs, or abandon it of course)

linawolf (approved):

brotkrueml (approved):

PR #76, opened 2023-11-15
[TASK] Script to update integration test baseline

linawolf:
Changes to the layout might affect many or all integration test files. In this case automatically update the tests to have a clean baseline again

### DIFF
--- a/tools/integration-test-baseline.php
+++ b/tools/integration-test-baseline.php
@@ -1,0 +1,45 @@
+<?php
+
+require_once __DIR__ . '/../vendor/autoload.php';
+
+copyTests(__DIR__ . '/../tests');
+
+function copyTests(string $directory): void
+{
+    $finder = new \Symfony\Component\Finder\Finder();
+    $finder
+        ->directories()
+        ->in($directory)
+        ->depth('== 0');
+
+    $tests = [];
+
+    foreach ($finder as $dir) {
+        if (!file_exists($dir->getPathname() . '/input')) {
+            copyTests($dir->getPathname());
+            continue;
+        }
+        $tempDir = $dir->getPathname() . '/temp';
+        $expectedDir = $dir->getPathname() . '/expected';
+        if (!file_exists($tempDir)) {
+            echo "Skipped $dir - temp directory does not exist\n";
+            continue;
+        }
+
+        $fileFinder = new \Symfony\Component\Finder\Finder();
+        $fileFinder
+            ->files()
+            ->in($dir->getPathname() . '/expected');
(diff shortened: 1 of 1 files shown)


tools/integration-test-baseline.php
```diff
@@ -0,0 +1,45 @@
...
+    $tests = [];
+
+    foreach ($finder as $dir) {
+        if (!file_exists($dir->getPathname() . '/input')) {
+            copyTests($dir->getPathname());
+            continue;
+        }
+        $tempDir = $dir->getPathname() . '/temp';
+        $expectedDir = $dir->getPathname() . '/expected';
+        if (!file_exists($tempDir)) {
```
garvinhicking, line 24: Should probably be "is_dir" instead to make the check cleaner?

I wonder though - how exactly does this work? Where does "temp" come from? It's not created from this script?!

garvinhicking (commented):
I like having a script like this and would opt to include it. We could spawn it from the Makefile as a "make integration-baseline" or something like that.

For me it's ok to have this as an "outside" PHP script (and not bash) if it's easier to maintain, and follows our main language experience.

Maybe one more echo at the start like:

```
echo "Updating baseline integration test results to match currently generated output."
```

or something like that?

tools/integration-test-baseline.php
linawolf: the temp is created from running the integration tests

garvinhicking: So the workflow to use this script would be:

* run all tests
* run `php tools/integration-test-baseline.php`?

Maybe then the error message could be updated to say "No temp directory exists, please run all tests (make pre-commit-tests) before executing this script" or something like that?

linawolf: I would just make the make command so that it first runs test-integration and then my script?

tools/integration-test-baseline.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+require_once __DIR__ . '/../vendor/autoload.php';
+
+copyTests(__DIR__ . '/../tests');
+
+function copyTests(string $directory): void
+{
+    $finder = new \Symfony\Component\Finder\Finder();
+    $finder
```
brotkrueml, line 10: I think that can be done easier:
```php
$finder
  ->directories()
  ->in($directory)
  ->name('input')
```

This way you don't need a recursion, but get all input dirs inside the tests folder at once.
  

https://symfony.com/doc/current/components/finder.html#file-name

linawolf:
This became obsolete with https://github.com/TYPO3-Documentation/render-guides/pull/82

Issue #78, opened 2023-11-16
Rendering "Documentation" merges contents of subdirectories into the main index


garvinhicking:
I'll just leave this here as a reminder so that I don't forget:

Rendering the Documentation via `vendor/bin/guides Documentation` currently produces a Index.html that looks off (it contains all content of all subdirectories placed in one page). Links inside that Document do not lead to the .html variants.

The created HTML-Files in the subdirectories look much more proper, however links seem to duplicate relative directory paths (leading to Developer/Developer/Index.html instead of Developer/Index.html).

Lina mentioned she'll have a look at this - let me know if it cannot be reproduced or what you find out :-)

garvinhicking:
Seems this is caused by the "single-html" rendering method. It currently would create a "index.html" and an "Index.html" file. The "Index.html" is the result of the "Index.rst" rendering, and "index.html" is created by the single-page rendering. Maybe this could be renamed to something like "singlepage-index.html" or so? Or needs to be put into a subdirectory like "_single" or so?

I only see the "index.html" because that is probably the last file created, and overwrites the proper "Index.html" in my MacOS setup.

PR #79, opened 2023-11-16
[BUGFIX] Creates distinct SinglePage index file (singlepage-index.html)

garvinhicking:
This prevents overwriting a "Index.html" file.

Resolves: #78

### DIFF
--- a/packages/typo3-guides-extension/src/Renderer/SinglePageRenderer.php
+++ b/packages/typo3-guides-extension/src/Renderer/SinglePageRenderer.php
@@ -25,7 +25,7 @@ final class SinglePageRenderer implements TypeRenderer
         );
 
         $context->getDestination()->put(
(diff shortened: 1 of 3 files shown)


packages/typo3-guides-extension/src/Renderer/SinglePageRenderer.php
```diff
@@ -25,7 +25,7 @@ public function render(RenderCommand $renderCommand): void
         );
 
         $context->getDestination()->put(
-            $renderCommand->getDestinationPath() . '/index.html',
+            $renderCommand->getDestinationPath() . '/singlepage-index.html',
```
linawolf, line 28: In the current rendering the file is saved at singlehtml/Index.html

I would suggest, we do the same

linawolf (approved):

linawolf:
Thanks for fixing this!

PR #82, opened 2023-11-16
[TASK] Split up integration tests in full tests and tests between markers

linawolf:


tests/Integration/IntegrationTest.php
```diff
@@ -41,19 +43,12 @@
 
 final class IntegrationTest extends ApplicationTestCase
 {
-    protected function setUp(): void
-    {
-        setlocale(LC_ALL, 'en_US.utf8');
-    }
+    private const CONTENT_START = '<!-- content start -->';
+    public const CONTENT_END = '<!-- content end -->';
```
brotkrueml, line 47:
```suggestion
    private const CONTENT_END = '<!-- content end -->';
```

PR #84, opened 2023-11-17
Use translate attribute

brotkrueml:
"notranslate" should avoid that a translation service like from Google and Microsoft
translates a code snippet. However, in HTML 5 there is a dedicated translate attribute
which is exactly for this case. Google and Microsoft respect this attribute, so
we switch to the standard instead of using a class.
    
Related: https://www.w3.org/International/questions/qa-translate-flag.en.html


linawolf (approved):

linawolf:
Thanks for this change!

PR #89, opened 2023-11-18
[BUGFIX] Do not shorten full integration tests

linawolf:
The integration-baseline script also shortened HTML files in the directory tests-full where the full files should be kept

Additionally do not copy .log files

### DIFF
--- a/tools/integration-test-baseline.php
+++ b/tools/integration-test-baseline.php
@@ -2,9 +2,10 @@
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
-copyTests(__DIR__ . '/../tests');
+copyTests(__DIR__ . '/../tests/Integration/tests');
+copyTests(__DIR__ . '/../tests/Integration/tests-full', false);
 
-function copyTests(string $directory): void
+function copyTests(string $directory, bool $shortenHtml = true): void
 {
     $finder = new \Symfony\Component\Finder\Finder();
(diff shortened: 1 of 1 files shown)


tools/integration-test-baseline.php
```diff
@@ -33,7 +34,9 @@ function copyTests(string $directory): void
             $tempFile = $tempDir . '/' . $fileName;
             $outputFile = $expectedDir . '/' . $fileName;
             if (file_exists($tempFile)) {
-                if (pathinfo($tempFile, PATHINFO_EXTENSION) === 'html') {
+                if (pathinfo($tempFile, PATHINFO_EXTENSION) === 'log') {
+                    echo sprintf("Ignoring log file %s\n", $outputFile);
```
brotkrueml, line 38:
```suggestion
                    echo sprintf("Ignoring log file %s\n", $outputFile);
                    continue;
```

I would like to see early `continue` also for the other cases to reduce cognitive complexity.

brotkrueml (changes requested):

tools/integration-test-baseline.php
linawolf: I just don't see the sense of using continue (which I find cognitive complex) when you can easily do without it?

linawolf: done, please have another look

PR #90, opened 2023-11-18
[TASK] Skip integration tests correctly

brotkrueml:
An integration test marked with the "skip" file should really be skipped. It should not be executed as the test may throw an error which is not desired (hence it it marked skipped).

### DIFF
--- a/tests/Integration/IntegrationTest.php
+++ b/tests/Integration/IntegrationTest.php
@@ -55,67 +55,61 @@ final class IntegrationTest extends ApplicationTestCase
         self::assertNotEmpty($compareFiles);
 
         $skip = file_exists($inputPath . '/skip');
+        if ($skip) {
+            self::markTestSkipped(file_get_contents($inputPath . '/skip') ?: '');
+        }
+
         $configurationFile = null;
         if (file_exists($inputPath . '/guides.xml')) {
             $configurationFile = $inputPath . '/guides.xml';
         }
 
-        try {
-            system('mkdir ' . escapeshellarg($outputPath));
+        system('mkdir ' . escapeshellarg($outputPath));
 
-            $this->prepareContainer($configurationFile);
-            $command = $this->getContainer()->get(Run::class);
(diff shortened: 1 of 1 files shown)


linawolf (changes requested):
Actually the behaviour when skipping tests is intended and served us well in the guides:

We write skipped tests for missing or wrong behaviour. then when a change changes the behaviour such as that the skipped test works now the skipped test will fail and inform you that it works now. This has popped up and informed us in many cases about skipped tests that work now.

Also if you do not which to execute a test you should not have it?

brotkrueml:
@linawolf 

I have currently the problem with the tests for code blocks: I want to document the current behaviour. So, one feature is not yet implemented, but a test should be there (which fails then of course). So, skipping (which means, don't do anything with it, at least when using phpunit directly), is the way to go (I thought).

But, as the test fails currently (the last assertFalse is failing) this cannot be merged as then the main branch would be red.

IMHO skip means skip: Don't do anything. But: Assertions are done also for skipped tests and the exception then catched and another assertion at the end. This is not skipping and fully unexpected knowing phpunit.

You are using the "skip" as "incomplete". So maybe we should implement an "incomplete" file which then marks the test as incomplete, like it is done now with the "skip" file.

We can implement another behaviour for that what I expect. Maybe adding an "ignore" file - which then marks the test as skipped. You now get my pain with the word, I hope.

linawolf:
I think if to phpunit "skip" means stop here and do nothing, then we should keep that vocabulary. I like your idea about introducing "incomplete" instead.

brotkrueml:
> I think if to phpunit "skip" means stop here and do nothing, then we should keep that vocabulary. I like your idea about introducing "incomplete" instead.

Marking a test as skipped throws an exception internally, which does not execute the code below the `markTestSkipped()` method. That exception is catched later and the `S` is displayed in the test output progress instead of `.`.

Will come up later with a updated PR.

brotkrueml:
Closed in favour of #91.

PR #106, opened 2023-11-20
[FEATURE] Implement converter for Settings.cfg to guides.xml

garvinhicking:


Documentation/guides.xml
```diff
@@ -4,6 +4,7 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
 >
+    <project title="Render guides" />
```
brotkrueml, line 7: I assume, this is doubled by accident?

composer.json
```diff
@@ -17,19 +17,21 @@
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-strict-rules": "^1.5",
         "phpunit/phpunit": "^10.4",
+        "symfony/console": "^6.3",
         "symplify/monorepo-builder": "^11.2",
         "symplify/phpstan-rules": "^12.4",
+        "t3docs/cli": "@dev",
```
brotkrueml, line 23: I am not sure that this is correct in the main composer.json as the other t3docs packages are not included here.

composer.json
```diff
@@ -17,19 +17,21 @@
...
+        "t3docs/cli": "self.version",
         "t3docs/php-domain": "self.version",
         "t3docs/typo3-docs-theme": "self.version",
         "t3docs/typo3-guides-extension": "self.version"
     },
-    "minimum-stability": "dev",
-    "prefer-stable": true,
     "autoload": {
         "psr-4": {
+            "T3Docs\\Cli\\": "packages/cli/src/",
```
brotkrueml, line 34: I suggest to rename the namespace, as t3docs/cli does not name the purpose correctly, perhaps something like t3docs/typo3-docs-theme-cli. What do you think?

packages/cli/bin/console
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env php
```
brotkrueml, line 1: We should rename the console name as this as available later as `vendor/bin/console` which is not distinctable between all the other commands. Perhaps something like `t3docs-theme`?

packages/cli/bin/console
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace T3Docs\Cli;
+
+use T3Docs\Cli\Command\Console;
+use Symfony\Component\Console\Application;
+
+$vendorDir = dirname(__DIR__) . '/../../vendor';
+$autoloadDirectory = $vendorDir . '/autoload.php';
```
brotkrueml, line 12: $autoloadDirectory is not the correct variable name as this holds a file path, not a directory path.

packages/cli/bin/console
```diff
@@ -0,0 +1,38 @@
...
+if (file_exists($autoloadDirectory)){
```
brotkrueml, line 13: Why do you think this is needed? If the autoload.php is not available in the vendor folder then also this application is not available? Or do I overlook something?

packages/cli/bin/console
```diff
@@ -0,0 +1,38 @@
...
+        if ($vendorDir === $rootPath) {
+            throw new \RuntimeException("Could not find autoload.php", 1636451408);
+        }
+        $vendorDir = \dirname($vendorDir);
+    }
+    require $vendorDir . '/autoload.php';
+}
+
+
+$application = new Application('cli', '0.0.1');
```
brotkrueml, line 32: 0.0.1 is no version number, as you provide now a bug fix release of a non-release (0.0.0.). But I suggest to avoid the version number as this has to be updated every time before tagging a new release to be in sync.

packages/cli/bin/console
```diff
@@ -0,0 +1,38 @@
...
+$command = new Console();
+
+$application->add($command);
+
+$application->setDefaultCommand($command->getName(), true);
```
brotkrueml, line 37: For now it is okay to set a default command, but when we get other commands we should remove it.

packages/cli/composer.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "t3docs/cli",
+    "description": "A collection of CLI binaries",
```
brotkrueml, line 3:
```suggestion
    "description": "A collection of CLI binaries for the TYPO3 documentation theme",
```

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+
+namespace T3Docs\Cli\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class Console extends Command
```
brotkrueml, line 13: I suggest to rename the class to make the purpose clearer: maybe MigrateSettingsCommand or similar.

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+        'html_theme_options',
+        'general',
+        'intersphinx_mapping'
+    ];
+
+    public function __construct(
+    ) {
+        parent::__construct('console');
+
+        $this->addArgument(
```
brotkrueml, line 65: Shouldn't the defining of the arguments in the `configure` method?
https://symfony.com/doc/current/console.html#configuring-the-command

I suggest also to add the desciption and the help text.

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+            'input',
+            InputArgument::REQUIRED,
+            'Path to directory where settings.cfg is stored, and guides.xml will be generated.',
+        );
+
+        $this->addOption(
+            'force',
+            null,
+            InputOption::VALUE_NONE,
+            'When set, forces to overwrite possibly existing XML.'
```
brotkrueml, line 75:
```suggestion
            'When set, forces the overwrite of a possibly existing guides.xml.'
```

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+    ];
+
+    public function __construct(
+    ) {
+        parent::__construct('console');
+
+        $this->addArgument(
+            'input',
+            InputArgument::REQUIRED,
+            'Path to directory where settings.cfg is stored, and guides.xml will be generated.',
```
brotkrueml, line 68:
```suggestion
            'Path to the "Documentation" directory where Settings.cfg is stored, and guides.xml will be generated.',
```

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $settingsFile = $input->getArgument('input') . '/Settings.cfg';
+        $guidesFile = $input->getArgument('input') . '/guides.xml';
+
+        if (!file_exists($settingsFile)) {
+            $output->writeln('<error>Could not open ' . $settingsFile . '</error>');
```
brotkrueml, line 86: As file_exists checks only the existing of a file, not if it is readable the message should be adapted, like `$settingsFile is not found` or similar.

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+            return Command::FAILURE;
+        }
+
+        if (file_exists($guidesFile) && !$input->getOption('force')) {
+            $output->writeln('<error>Target file already exists in specified directory (' . $guidesFile . ')</error>');
+            return Command::FAILURE;
+        }
+
+        $output->writeln('Converting ' . $settingsFile . ' to ' . $guidesFile . ' ...');
```
brotkrueml, line 95:
```suggestion
        $output->writeln('Migrating ' . $settingsFile . ' to ' . $guidesFile . ' ...');
```

should be a better word.

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+
+        if (!$this->convertSettingsToGuide($output, $settingsFile, $guidesFile)) {
+            $output->writeln('<error>Settings could not be converted. Please check for proper syntax.</error>');
+            return Command::FAILURE;
+        }
+
+        $output->writeln('Settings converted. You can now delete Settings.cfg and add guides.xml to your repository.');
```
brotkrueml, line 102: What if we just rename Settings.cfg to Settings.cfg.bak. This way it should no longer be in the version control system.

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+
+        // This array will hold all setting values that we could not migrate to guides.xml
+        $unmigratedSettings = $settings;
+
+        // Holds number of converted settings
+        $convertedSettings = 0;
+
+        // Setup basic DOM Document
+        $dom = new \DOMDocument("1.0", "UTF-8");
+        $dom->preserveWhiteSpace = false;
```
brotkrueml, line 128: That makes reading the XML file by humans harder, as all tags are in one line?

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+        $dom->formatOutput = true;
+
+        // Add static <guides> element with proper XMLNS
+        $guides = $dom->createElement('guides');
+        $guides->setAttribute('xmlns', 'https://www.phpdoc.org/guides');
+        $guides->setAttribute('xmlns:xsi', 'http://www.w3.org/2001/XMLSchema-instance');
+        $guides->setAttribute('xsi:schemaLocation', 'https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd');
```
brotkrueml, line 135:
```suggestion
        $guides->setAttribute('xsi:schemaLocation', 'https://www.phpdoc.org/guides ../vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd');
```

As it is available in Documentation folder, so one level up. But this may also not work, many extensions define a .Build/vendor folder. But as we don't know this, that is fine for now. Maybe we should hint a message about it at the end.

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+        $guides->setAttribute('links-are-relative', 'true');
+
+        // Add <extension> Element. This gets filled with the old "html_theme_options" section.
+        $extension = $dom->createElement('extension');
+        $extension->setAttribute('class', '\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension');
+        if (isset($settings['html_theme_options']) && is_array($settings['html_theme_options'])) {
```
brotkrueml, line 141:
```suggestion
        if (is_array($settings['html_theme_options'] ?? false)) {
```

should be enough, I think?

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+            $output->writeln('<error>Settings could not be converted. Please check for proper syntax.</error>');
+            return Command::FAILURE;
+        }
+
+        $output->writeln('Settings converted. You can now delete Settings.cfg and add guides.xml to your repository.');
+
+        return Command::SUCCESS;
+    }
+
+    protected function convertSettingsToGuide(OutputInterface $output, string $inputFile, string $outputFile): bool
```
brotkrueml, line 107: Personally I would split up this long method for better readability:
- createDomDocument()
- addExtensionSettings()
- addProjectSettings()
- addInterlinks()

So this method only serves as a starting point which then calls the according methods.

packages/cli/src/Command/Console.php
```diff
@@ -0,0 +1,225 @@
...
+
+        $fp = fopen($outputFile, 'w');
+        if (!$fp) {
+            $output->writeln('<error>Could not create file ' . $outputFile . '</error>');
+            return false;
+        }
+
+        fwrite($fp, (string)$dom->saveXML());
+
+        $output->writeln('<info>' . $convertedSettings . ' Settings were migrated.</info>');
```
brotkrueml, line 221:
```suggestion
        $output->writeln('<info>' . $convertedSettings . ' settings were migrated.</info>');
```

brotkrueml (changes requested):

brotkrueml:
Additionally, CGL must be followed ;-)

Documentation/guides.xml
garvinhicking: Yes, accidentall commit sadly. Reverted that part. Thanks for catching this :)

packages/cli/bin/console
garvinhicking: (We have the same code in the current bin/guides-cli, will check how this can be adapted)

packages/cli/src/Command/Console.php
garvinhicking: Changed this to `true` - however, the XML document is created from a empty DomDocument and not reading in an existing XML file, so there's no real whitespace to preserve.

packages/cli/bin/console
garvinhicking: (This mainly stems from the reason above, I copy+pasted it from the bin/guides runner)

packages/typo3-guides-cli/README.rst
```diff
@@ -0,0 +1,11 @@
+
+===========================================
+Toolbox for render-guides CLI operations
+===========================================
+
+You can also require this package in your composer.json::
+
+    composer req t3docs/guidescli
```
brotkrueml, line 8:
```suggestion
    composer req t3docs/typo3-guides-cli
```

packages/typo3-guides-cli/src/Command/MigrateSettingsCommand.php
```diff
@@ -0,0 +1,306 @@
...
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class MigrateSettingsCommand extends Command
+{
+    protected static $defaultName = 'typo3-guides';
```
brotkrueml, line 15: Class is final, so use private please ;-) Also in the methods below.

brotkrueml (changes requested):

packages/typo3-guides-cli/src/Command/MigrateSettingsCommand.php
garvinhicking: Changed this, where it worked. Note that $defaultName MUST be protected:

```
PHP Fatal error:  Access level to T3Docs\GuidesCli\Command\MigrateSettingsCommand::$defaultName must be protected (as in class Symfony\Component\Console\Command\Command) or weaker in /Users/garvin/TYPO3/render-guides/packages/typo3-guides-cli/src/Command/MigrateSettingsCommand.php on line 13
```

packages/typo3-guides-cli/src/Command/MigrateSettingsCommand.php
```diff
@@ -0,0 +1,306 @@
...
+    /**
+     * Setup basic DOM Document
+     **/
+    private function createXmlSectionGuides(): \DOMElement
+    {
+        // Add static <guides> element with proper XMLNS
+        $guides = $this->xmlDocument->createElement('guides');
+        $guides->setAttribute('xmlns', 'https://www.phpdoc.org/guides');
+        $guides->setAttribute('xmlns:xsi', 'http://www.w3.org/2001/XMLSchema-instance');
+        $guides->setAttribute('xsi:schemaLocation', 'https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd');
```
brotkrueml, line 140:
```suggestion
        $guides->setAttribute('xsi:schemaLocation', 'https://www.phpdoc.org/guides ../vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd');
```

This way the xsd is found.

brotkrueml (commented):
Personally, I don't like that you added two other commands to this one, as that is not in the scope of the converter.

packages/typo3-guides-cli/src/Command/MigrateSettingsCommand.php
garvinhicking: This will never properly resolve, because guides.xml might be in the root, or in other subdirectories.

At least for the root guides.xml this is aimed, there "../" would break it...

garvinhicking: Or wait, if its relative to the URI it might work and I misunderstood it. Will investigate.

garvinhicking: Adjusted to be relative after discussion :-)

brotkrueml (approved):

Issue #83, opened 2023-11-16
Documentation/guides.xml is ignored


linawolf:
When I place intersphinx links in the file Documentation/guides.xml the are ignored. Only if they are in the main dirctory they are interpreted.

garvinhicking:
@linawolf How exactly does your guides.xml look like? How to you generate the docs?

I believe the way it's currently meant to work is that you have to specify the guides.xml to be loaded specifically:

````
vendor/bin/guides --config Documentation/ Documentation
````

This would first load the root guides.xml (see line 37 in that binary). And then, it would overload/merge it with the specified one.

So, either we would need to always intentionally use the guides.xml of a subdirectory through the command.

I suppose we could change the `RunDecorator.php` to also let this fall back on the 'guessInput' return through a check like this:

````
        if (!empty($guessedInput['input'])) {
            // Don't know how to do this here though; don't know how/where to access $containerFactory,
            // or if it's too late already
            $containerFactory->addConfigFile($input->getParameterOption('--config', getcwd(), true).'/guides.xml');
        }
````

See the inline comment though were I fail, and where we'd probably need feedback from Jaap, if we want to follow up on that approach.

**Comment by linawolf** [comment]
So, in the publishing process, as far as I understand it, we check if there is a file Documentation/guides.xml and if so we start the the render-guides instead of sphinx, However in that process it seems like the Documentation/guides.xml is being ignored but the script uses a guides.xml in the root  directory of the file only

**Comment by garvinhicking** [comment]
Hhm. Which publishing process do you mean exactly? The github actions in a seperate directory? Maybe there this "--config" parameter would need to be added?!

**Comment by linawolf** [comment]
That would be in https://github.com/TYPO3-Documentation/gh-render-action

**Comment by garvinhicking** [comment]
Interesting. Rabbitholy :) 
Seems that this:

```
uses: TYPO3-Documentation/render-guides@main
      id: render-guides
      if: steps.enable_guides.outputs.GUIDES == 'true'
      with:
        config: t3docsproject/Documentation/
        output: t3docsproject/RenderedDocumentation/Result/project/0.0.0
```

Actually specifies the "config" option. I'm not proficient enough to know how this could be debugged, how I can locally "emulate" github actions to see which commands get executed here...

garvinhicking:
Maybe a dumb idea to debug this, but maybe we could:

- create a "run.log" through the guides binary, that shows all used config options and project settings
[...]

PR #119, opened 2023-11-22
[TASK] Increases verbose output when debugging (-vvv) is set

garvinhicking:
To debug https://github.com/TYPO3-Documentation/render-guides/issues/83 we need more verbose output what directories/files are used.

This depends on https://github.com/phpDocumentor/guides/pull/705


brotkrueml (commented):
We should then also decrease the `-vvv` when running `make docs` to `-vv` to avoid that much noise when running the guides command.

garvinhicking:
> We should then also decrease the `-vvv` when running `make docs` to `-vv` to avoid that much noise when running the guides command.

Yes - however, we'll first need to have it enabled, because this debugging is required to see why the github-render-action fails to properly configure a project. Or we need a way to specify "-vvv" for a github action, but a problem is that I can only see how to add options with two dashes to a github action and a specifiy name, not with a single dash - see https://github.com/phpDocumentor/guides/pull/705#issuecomment-1822936243

brotkrueml:
> > We should then also decrease the `-vvv` when running `make docs` to `-vv` to avoid that much noise when running the guides command.
> 
> Yes - however, we'll first need to have it enabled, because this debugging is required to see why the github-render-action fails to properly configure a project. Or we need a way to specify "-vvv" for a github action, but a problem is that I can only see how to add options with two dashes to a github action and a specifiy name, not with a single dash - see [phpDocumentor/guides#705 (comment)](https://github.com/phpDocumentor/guides/pull/705#issuecomment-1822936243)

Well, instead of using "make docs" in the action you can use the full command with "-vvv" (also for locally debugging). Debug information which is not needed locally on rendering most of the time can make warnings hard to recognise.

garvinhicking:
> Well, instead of using "make docs" in the action you can use the full command with "-vvv" (also for locally debugging). Debug information which is not needed locally on rendering most of the time can make warnings hard to recognise.
> 
> Or add a "make docs-debug" option for convenience.

Please check out the YAML of the github action. There's no make file being run, but an entrypoint of the docker image, and arguments are passed differently... not sure how to proceed here.

brotkrueml:
> Please check out the YAML of the github action. There's no make file being run, but an entrypoint of the docker image, and arguments are passed differently... not sure how to proceed here.

Yup, but that does not invalidate my remark about the Makefile:
https://github.com/TYPO3-Documentation/render-guides/blob/main/Makefile#L67

garvinhicking:
> Yup, but that does not invalidate my remark about the Makefile: https://github.com/TYPO3-Documentation/render-guides/blob/main/Makefile#L67

Just to flesh this out:

* First we need to figure out a way to add toggle-able debug output to https://github.com/phpDocumentor/guides/pull/705 (I don't know how)
* Once that is done, we need to merge both PRs
* Then we need to run the github action and see why the rendering fails to recognize guides.xml (which needs the -vvv of the makefile)
* Once we figured this out, we remove the extra verbosity from the command of this repository (Makefile) (and may need to re-enable it if we face further issues with the docker image)

The best way of course would be to be able to pass verbosity commands to the github workflow, then we could directly remove the "-vvv" from the default Makefile command.

Looking forward to ideas/input/help, I believe Jaapio offered to check into it when there's time!

brotkrueml (approved):

brotkrueml:
As the adaption of the Makefile can also be done later, approved.

PR #121, opened 2023-11-22
[FEATURE] Introduce Textroles t3ext and t3src

linawolf:


garvinhicking (changes requested):

linawolf:
We introduced a new variable for the currently supported TYPO3 version so that the links go on the currently supported version. We will also use it in different scenarios

garvinhicking (approved):

PR #124, opened 2023-11-23
[TASK] Use t3docs/guides-php-domain

brotkrueml:


composer.json
```diff
@@ -24,14 +25,20 @@
...
     },
     "replace": {
-        "t3docs/php-domain": "self.version",
         "t3docs/typo3-docs-theme": "self.version",
         "t3docs/typo3-guides-extension": "self.version"
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/brotkrueml/twig-codehighlight"
```
garvinhicking, line 34: Could you add your package on packagist, so that a distinct repository wouldn't be needed?

garvinhicking (approved):

composer.json
brotkrueml: Yup, as far as I have convinced myself it should be ;-)

garvinhicking: I give you a magic +1 Stick of Conviction! :-D

linawolf (approved):

PR #125, opened 2023-11-23
[TASK] Use absolute URI for XSD schema

garvinhicking:
(See proposal in Slack channel)


brotkrueml:
As already communicated in Slack, my -1.

Additionally, we should not link to a raw.github.... URL but use a URL we control.

PR #126, opened 2023-11-23
[TASK] Renames command to "lint-guides-xml" and utilize it in workflow

garvinhicking:
Removes xmllint.sh, adapts Makefile for "make test-xml"


Makefile
```diff
@@ -131,7 +131,7 @@ test-unit: ## Runs unit tests with phpunit
 
 .PHONY: test-xml
 test-xml: ## Lint all guides.xml
-	./tools/xmllint.sh
+	$(PHP_BIN) packages/typo3-guides-cli/bin/typo3-guides lint-guides-xml
```
garvinhicking, line 134: Note: We'll have to adapt this, once we can somehow get this binary to be put inside ./bin/typo3-guides parallel to the existing `guides` binary.

linawolf (approved):

Makefile
brotkrueml: Nope, I don't think so. The Makefile is only available in the monorepo easing our work, is will not be available in the splitted repos. So this is valid and we don't have to think later about this.

.github/workflows/main.yaml
```diff
@@ -41,13 +41,8 @@ jobs:
...
         run: "make test-unit ENV=${{ env.RUN_ENVIRONMENT }}"
 
-      - name: "Install libxml2-utils"
-        run: |
-          sudo apt-get update
-          sudo apt -y install libxml2-utils
-
-      - name: "Lint xml configs"
-        run: "tools/xmllint.sh"
+      - name: "Lint guides.xml configs"
```
brotkrueml, line 44:
```suggestion
      - name: "Lint guides.xml configurations"
```

brotkrueml (approved):

Makefile
garvinhicking: Sorry, I can't follow, why would the splitted repos perform any workflows that aren't performed in the monorepo actually? I always thought a split repository does nothing on its own, the monorepo pushes to the split ones only?

brotkrueml: When I run "make docs" locally on the monorepo I will always see all the debug information I don't want and I don't need. But anyway, I am not using the make task for generating the docs ;-) Just my 2 cents.

PR #130, opened 2023-11-23
[FEATURE] Add code highlighting to theme

brotkrueml:


brotkrueml:
Currently throws an error on rendering:

```
In layout.html.twig line 18:
                                                                                                                                   
  An exception has been thrown during the rendering of a template ("Cannot get current file name when not rendering a document").  
                                                                                                                                   

In RenderContext.php line 146:
                                                              
  Cannot get current file name when not rendering a document  
```

linawolf:
@jaapio getRelativePaths is only calculating the paths, the files are moved in the post rendering event

linawolf:
We need https://github.com/phpDocumentor/guides/pull/721 to calculate the relative paths

linawolf:
@brotkrueml with the latest changes to the guides the Documentation in the "Documentation"´folder now renders without errors or warnings

PR #142, opened 2023-11-25
[TASK] Move logic for writing guides.xml into a separate class

brotkrueml:
Additionally, require "ext-dom" in the root composer.json to avoid complaints of missing extension in PhpStorm.

### DIFF
--- a/packages/typo3-guides-cli/src/Repository/Exception/FileException.php
+++ b/packages/typo3-guides-cli/src/Repository/Exception/FileException.php
@@ -26,6 +26,16 @@ final class FileException extends \RuntimeException
         );
     }
 
+    public static function notWritable(string $filePath): self
+    {
+        return new self(
+            \sprintf(
+                'File "%s" cannot be written!',
+                $filePath,
+            )
+        );
+    }
+
     public static function notParsable(string $filePath): self
     {
         return new self(
--- a/packages/typo3-guides-cli/src/Repository/GuidesSettingsRepository.php
+++ b/packages/typo3-guides-cli/src/Repository/GuidesSettingsRepository.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace T3Docs\GuidesCli\Repository;
+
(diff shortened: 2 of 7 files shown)


packages/typo3-guides-cli/src/Repository/Exception/FileException.php
```diff
@@ -26,6 +26,16 @@ public static function notReadable(string $filePath): self
         );
     }
 
+    public static function notWritable(string $filePath): self
```
linawolf, line 29: Do we have to introduce our own exceptions for this? It seems like a common issue to me?

packages/typo3-guides-cli/src/Repository/GuidesSettingsRepository.php
```diff
@@ -0,0 +1,22 @@
...
+class GuidesSettingsRepository
+{
+    /**
+     * @return int Number of bytes written
+     */
+    public function create(string $filePath, \DomDocument $document): int
+    {
+        // An error is suppressed to avoid PHP warnings when the file does
+        // not exist. As we check the return value later this should be okay.
+        $result = @file_put_contents($filePath, $document->saveXml());
```
linawolf, line 18: Can't we use some library for writing files that offers all this?

packages/typo3-guides-cli/src/Repository/Exception/FileException.php
brotkrueml: I like to move the instantiation of an exception into the exception itself (with named constructors). This way it is cleaner and the throwing class does not have to deal with wordings.

packages/typo3-guides-cli/src/Repository/GuidesSettingsRepository.php
brotkrueml: Hmm, I will postpone this PR and handle the XML stuff first (which is far more complex and with more code in the command). Then we will see, if this PR is necessary in that way.

brotkrueml:
Closed in favour of #251.

PR #143, opened 2023-11-25
[TASK] Update codehighlight extension to log non-existing languages

brotkrueml:
The logger will be injected automatically by dependency injection.

A log entry looks now like this:

    [2023-11-25T19:34:34.186907+01:00] app.WARNING: Language "kjsdfhs" is not available to highlight code [] []


linawolf:
could you add an integration test for that?

linawolf (approved):

PR #144, opened 2023-11-26
[TASK] Create a make command for settings migration

linawolf:
And add some documentation


Documentation/Migration/Index.rst
```diff
@@ -0,0 +1,29 @@
...
+..  code-block:: sh
+
+    make migrate-settings path=path/to/Documentation
+
+With ddev
+=========
+
+..  code-block:: sh
+
+    ddev exec packages/typo3-guides-cli/bin/typo3-guides migrate path/to/Documentation
```
garvinhicking, line 29: I'd prefer to advertise using either the makefile syntax or composer wrapper only, not introducint native calls to? So with either

ddev exec "make ENV=local migrate-settings ..."

Or

ddev composer make migrate-settings

(The other docs use the latter which is IMO clearer to read because it doesnt need the ENV=local part)

What do you think?

linawolf: You are right, changed it

Documentation/Migration/Index.rst
```diff
@@ -0,0 +1,29 @@
...
+=========
+Migration
+=========
+
+If you want to migrate from using Sphinx to render TYPO3 documentation to
+using the render-guides, the file :file:`Documentation/Settings.cfg` has to be
+replaced by an XML file, :file:`Documentation/guides.xml`.
+
+To facilitate migration the extension `t3docs/typo3-guides-cli` in this
+mono-repository comes with a Symfony Console Command to automatically migrate
```
brotkrueml, line 14:
```suggestion
mono-repository comes with a Symfony console command to automatically migrate
```

Documentation/Migration/Index.rst
```diff
@@ -0,0 +1,29 @@
...
+the outdated :file:`Settings.cfg`.
+
+With make
```
brotkrueml, line 17: And how do I use it without having make and ddev on my machine? We should also communicate how to do it with "raw" PHP.

Documentation/Migration/Index.rst
```diff
@@ -0,0 +1,29 @@
...
+..  code-block:: sh
+
+    make migrate-settings path=path/to/Documentation
+
+With ddev
+=========
+
+..  code-block:: sh
+
+    ddev exec ddev composer make migrate-settings path=path/to/Documentation
```
garvinhicking, line 29: Justd "ddev composer make..." - no "ddev exec" in front (sorry, don't k ow how to edit this directly on mobile)

brotkrueml, line 29: Remember, later people will require "t3docs/guides-cli" into their extension for the commands. This means, they have to configure a ddev box for their extension, and "build-essential" has to be added manually. So just giving the ddev command won't work for them.

Also this requirement is not mentioned above.

linawolf: This documentation is only for us that use the mono-repository. We can adjust it when we transfer it to the how-to-document to target a wider audience

Documentation/Migration/Index.rst
linawolf: If you know please add that information. As I cannot do it I cannot add that information

Documentation/Migration/Index.rst
```diff
@@ -0,0 +1,29 @@
...
```
linawolf, line 29:
```suggestion
    ddev composer make migrate-settings path=path/to/Documentation
```

Documentation/Migration/Index.rst
brotkrueml: I am using just the plain ol' PHP ;-) But as I am the minority, go ahead.

Documentation/Migration/Index.rst
brotkrueml: Done

brotkrueml (approved):

PR #146, opened 2023-11-26
[BUGFIX] Limit PHP to 8.1 during phar building.

linawolf:
shivammathur/setup-php@v2 currently fails with PHP 8.3 See for example https://github.com/TYPO3-Documentation/render-guides/actions/runs/6995451358/job/19030317471?pr=144

### DIFF
--- a/.github/workflows/phar.yaml
+++ b/.github/workflows/phar.yaml
@@ -7,6 +7,10 @@ on:
     types:
       - published
 
+env:
+  DEFAULT_PHP_VERSION: "8.1"
+  RUN_ENVIRONMENT: "local"
+
 name: Phar
 jobs:
   build:


garvinhicking:
Suggestions:
* use php 8.1
* enforce using 8.1 also for composer by including  "platform" setting?

brotkrueml:
> Suggestions:
> 
> * use php 8.1
> * enforce using 8.1 also for composer by including  "platform" setting?

Nope, we should not use the platform setting. If you want to enforce it require PHP with "~8.1.0".

brotkrueml (approved):
I am wondering, why this worked in the past, where the env variables were not set but used.

linawolf:
I suppose that shivammathur/setup-php defaults to the newest PHP version if none is set...

garvinhicking:
> > Suggestions:
> 
> > 
> 
> > * use php 8.1
> 
> > * enforce using 8.1 also for composer by including  "platform" setting?
> 
> 
> 
> Nope, we should not use the platform setting. If you want to enforce it require PHP with "~8.1.0". But I do not see a need for that, as we define it in the build which version.


That will prevent us from running composer in our local dev environments then, because I for example am running PHP 8.3 locally, the  I won't be able to run composer install anymore...

Of course we can do that, but then we lock contributors into ddev or docker mandatory, where the platform attribute (I know about its dangers) would still allow it locally...?

PR #148, opened 2023-11-26
[TASK] Move class constants into dedicated enums

brotkrueml:
The goal is to reduce the "god" command for migrating the legacy configuration. This change moves the class constants into dedicated enums to decouple them from the command. A next step would be to move the logic for building the XML into a dedicated class.


garvinhicking (approved):
Great. Good use of enums :-)

linawolf (approved):

PR #150, opened 2023-11-27
[TASK] Move XML handling out of MigrateSettingsCommand

brotkrueml:
The return value of the migrate() method is still ugly, but can be addressed later in a separate change.

There are some glitches, like when a legacy setting has an empty value, a message "Some of your settings could not be converted" is displayed. Or if an id or value has whitespaces. This will also be addressed in a separate change to be as near as possible to the original code.

Additionally, the generating of the XML is now covered by unit tests.


garvinhicking (approved):
👍

linawolf (approved):
Nice work

PR #152, opened 2023-11-28
[TASK] Use php-cs-fixer directly

brotkrueml:
This change removes the package "typo3/coding-standards" in favour of "friendsofphp/php-cs-fixer". There is no need to use this proxy package.

Instead, we define the needed rules directly. As we stick zu the PER-CS standard (the successor of PSR-12) we pin the rules to the PER-CS 1 ruleset. Additionally, we opt in for some already implemented PER-CS 2 rules to avoid too much style changes already available in the codebase.

With the pinning to defined rules we ensure that updates of php-cs-fixer does not "break" the currently used styles locally or in CI. Changes in the defined rules can therefore be done explicitly.

### DIFF
--- a/packages/typo3-guides-cli/tests/unit/Migration/SettingsMigratorTest.php
+++ b/packages/typo3-guides-cli/tests/unit/Migration/SettingsMigratorTest.php
@@ -35,9 +35,9 @@ final class SettingsMigratorTest extends TestCase
         yield 'with empty legacy settings' => [
             'legacy settings' => [],
             'expected' => <<<EXPECTED
-<?xml version="1.0" encoding="UTF-8"?>
-<guides xmlns="https://www.phpdoc.org/guides" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://www.phpdoc.org/guides ../vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd" links-are-relative="true"/>
-EXPECTED,
+                <?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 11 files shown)


packages/typo3-guides-cli/tests/unit/Migration/SettingsMigratorTest.php
```diff
@@ -35,9 +35,9 @@ public static function providerForMigrateReturnsXmlDocumentCorrectly(): \Generat
         yield 'with empty legacy settings' => [
             'legacy settings' => [],
             'expected' => <<<EXPECTED
-<?xml version="1.0" encoding="UTF-8"?>
-<guides xmlns="https://www.phpdoc.org/guides" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://www.phpdoc.org/guides ../vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd" links-are-relative="true"/>
-EXPECTED,
+                <?xml version="1.0" encoding="UTF-8"?>
```
garvinhicking, line 38: Doesn't this change the expected output due to introduced whitespace?

brotkrueml: Nope, as we assert XML string with XML string in the unit tests.

brotkrueml:
@linawolf @garvinhicking Any more comments? Would like to merge this soon as this PR tends to be in conflict with new merges.

garvinhicking (approved):
Sorry, just got to it now. Since the   XML indentation shouldn't cause issues like you explained, please go ahead!

PR #160, opened 2023-12-08
[BUGFIX] Correct spelling of admonition scss file

brotkrueml:


garvinhicking (approved):
Good catch! Does that need to be changed in some other "source of truth" repository, did benji maybe migrate that from some other core Repo?

brotkrueml:
> Good catch! Does that need to be changed in some other "source of truth" repository, did benji maybe migrate that from some other core Repo?

See #161 ;-)

PR #161, opened 2023-12-08
[TASK] Link local resources

brotkrueml:
Local resources are linked instead of the ones from the Python-based theme. This is necessary for the next steps, for example, adjust the styles and JavaScripts to our needs. The formerly "_assets" folder was renamed to "_resources" to match the folder in the theme package. Additionally, the resources can now be copied recursively.

The styles for code highlighting are now embedded directly into the codeblock.css file. This file was not linked before, this is now added. Note: The code block has now a dark background color (as for the Python-based rendering). This will be adjusted in a follow-up.


garvinhicking (commented):
That looks helpful and I think it will be needed. However I can't tell about the implications, i.e. if the github ci-deploy action properly transfers all the local assets so that they can be used in the final rendered setup (and where exactly the "upload to cloud" step takes place, and how local assets would then be referenced to? Or where/how their location URIs are then changed to the cloud URI...

So bottom line: Proceed as you deem fit, I cannot really review this - and we'll "fix it in post" when things break. Did you ask Benji about this, he probably knows more about the cloud storage URI things?

brotkrueml:
> That looks helpful and I think it will be needed. However I can't tell about the implications, i.e. if the github ci-deploy action properly transfers all the local assets so that they can be used in the final rendered setup (and where exactly the "upload to cloud" step takes place, and how local assets would then be referenced to? Or where/how their location URIs are then changed to the cloud URI...
> 
> So bottom line: Proceed as you deem fit, I cannot really review this - and we'll "fix it in post" when things break. Did you ask Benji about this, he probably knows more about the cloud storage URI things?

The secrets are available in the respository already, so we can upload the resources in principle. But AFAIK this has to be implemented. As I am not that GitHub Actions guy, I can't tell more.

PR #162, opened 2023-12-08
[TASK] Style code blocks

brotkrueml:
Remove the superfluous pygments styles (they were used in the Python-based code highlighter) and define the code block styles based on the light theme. The "data-cbs-theme" attribute at the body tag was used for the dark/light mode switch for code blocks. This will be done better with a media query later.

Currently, there is a flaw in the font size. The font size is also not correct for the overall content, this will be tackled in a separate PR. Also, the code lines have a white background. This will also be adjusted in a consecutive PR.

We also have a conflict with text roles (`:css:`, `:php:`, etc.) as these are applied via `code .css`, `code .php` (which are then also used for the code block). So the colors are overwritten from the highlight theme. That's a reason why it is bad to style via tags and not via classes. This will be adjusted in a separate PR.


fe-hicking (approved):
I can't comment much on the implications of what styling gets changed/removed, but at least I see no obvious issues with this 😅

linawolf (approved):

linawolf:
@brotkrueml please rebase so we cna continue

brotkrueml:
> @brotkrueml please rebase so we cna continue

@linawolf Done.

PR #165, opened 2023-12-09
[TASK] Adjust classes for various code textroles

brotkrueml:
Previously, the styles were applies to the code tag with classes. This has the drawback that this infiltrates all other code tag styles which is undesired. Therefore, a dedicated class "code-inline" has been introduced to mitigate this.

Additionally, the previous "codeblock.scss" was renamed to "_code.scss" and integrated into the theme.scss (as the styles has to be loaded anyway for every page as there can be code textroles or code block directives).


linawolf (approved):

linawolf:
I will make a PR for the unkown -> unknown template. We have to change it then in the render-guides in a follow up

packages/typo3-docs-theme/resources/template/inline/textroles/unkown.html.twig
```diff
@@ -0,0 +1,3 @@
+{% apply spaceless %}
```
linawolf, line 1: Please make a composer update then rename this template to unknown.html.twig. See https://github.com/phpDocumentor/guides/pull/744

linawolf (changes requested):

packages/typo3-docs-theme/resources/template/inline/textroles/unkown.html.twig
brotkrueml: Done

linawolf (approved):

PR #167, opened 2023-12-10
[BUGFIX] Do not overwrite background color of code line in code block

brotkrueml:


packages/typo3-docs-theme/assets/sass/_highlightjs_github.scss
```diff
@@ -11,7 +11,7 @@
 
 .hljs {
   color: #24292e;
-  background: #ffffff;
+  //background: #ffffff; // The code-block has already a gray background set!
```
linawolf, line 14: Many style checkers for scss dissallow comented out code..

brotkrueml: Ours not. I want to keep it in case we update that file (it is from highlightjs) and have a comment about it. That way it is not added again then.

linawolf (approved):

brotkrueml:
Thanks. Another solution I considered was to use "background: transparent", but this is superfluous code. Also overwriting the background in the _code.scss file again. We will see, if the current solution is usable on updates of that file.

PR #168, opened 2023-12-11
[TASK] Add `generator` meta tag

garvinhicking:
Allows to easily differentiate between sphinx and phpdocumentor


packages/typo3-docs-theme/resources/template/structure/layout.html.twig
```diff
@@ -11,6 +11,7 @@
     <link href="http://purl.org/dc/elements/1.1/" rel="schema.dc">
     <meta content="{{ env.projectNode.lastRendered|date("Y-m-d\\TH:i:sP") }}" name="dc.modified">
     <meta content="{{ env.projectNode.lastRendered|date("Y-m-d\\TH:i:sP") }}" property="article:modified_time">
+    <meta content="phpdocumentor" name="docsearch:engine">
```
brotkrueml, line 14: I suggest to move it up some lines so it is together with the other docsearch meta tags. This makes it easier to grasp all of them at once.

garvinhicking: As discussed, I shuffled the order and put it to the top before the group of docsearch attributes. And now we use the established "generator" instead of a custom "docsearch:engine"

brotkrueml (approved):

Issue #158, opened 2023-12-08
RFC: Dynamic guides.xml adaptation for GitHub rendering action


garvinhicking:
# Current situation

## gh-render-action

We have a GitHub action at https://github.com/TYPO3-Documentation/gh-render-action/blob/main/action.yml which provides a GitHub action step to render documentation.

## t3docs-ci-deploy

That action is referred to in the Repository https://github.com/TYPO3-Documentation/t3docs-ci-deploy/blob/main/.github/workflows/main-rendering.yml:

```
      - name: Render Repository
        uses: TYPO3-Documentation/gh-render-action@main
        id: rendering
        with:
          repository_url: ${{ github.event.client_payload.repository_url }}
          source_branch: ${{ github.event.client_payload.source_branch }}
          target_branch_directory: ${{ github.event.client_payload.target_branch_directory }}
```

The action takes 3 arguments (repository url, source branch, target branch directory for the deployment).

Whenever the rendering of a repository like https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/tree/documentation-draft/Documentation is requested, this action gets executed with the payload input that specifies:

* The URL of the GIT repository to render (i.e. `https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi.git`)
* The GIT branch (i.e. `main` or `documentation-draft`) of the target repository to render
* The destination directory of the rendered documentation (i.e. `draft` - the part of `https://docs.typo3.org/m/typo3/reference-coreapi/***draft***/en-us`)

## back to gh-render-action

Those three parameters are specified in the trigger, get passed on to `gh-render-action` and prepares the local directory:

```
    - uses: TYPO3-Documentation/render-guides@main
      id: render-guides
      if: steps.enable_guides.outputs.GUIDES == 'true'
      with:
        working-directory: t3docsproject
        config: ./t3docsproject/Documentation/
        output: ./RenderedDocumentation/Result/project/0.0.0
```

The output directory is just temporary, the rest of the action actually takes care it gets put into the specified `${{ env.target_branch_directory }}`  directory tree.

**Please note**, at this point neither the variable `${{ env.source_branch }}` nor `${{ env.target_branch_directory }}` is passed (and evaluated) in the action `TYPO3-Documentation/render-guides@main`.

## render-guides GitHub action

[...]

PR #169, opened 2023-12-12
[TASK] Splits GitHub rendering action into composite

garvinhicking:
This PR splits the current process of single-use rendering up into a two-step composite process as outlined in  https://github.com/TYPO3-Documentation/render-guides/issues/158

The first step takes care of configuring a local guides.xml (during rendering/GH-action only) with dynamic parameters. Those parameters fixate the input from the GitHub action for the target GIT branch and target deployment directory.

The second step is the usual rendering, which will then take into account the modified parameters, so that the rendered output includes also the non-committed changes, as the old sphinx jobfile did.

The commit includes a large addition to the `ConfigureCommand` that is needed to configure the required attributes. That code will need later refactoring and added unit testing. For now it should work to progress on the GitHub action.

This PR should be merged in conjunction with the changes made to the `gh-render-action` PR: https://github.com/TYPO3-Documentation/gh-render-action/pull/21

### DIFF
--- a/.github/actions/configure-guides-step/action.yaml
+++ b/.github/actions/configure-guides-step/action.yaml
@@ -31,7 +31,3 @@ runs:
     - '--extension-attribute="edit-on-github-branch"'
     - '--extension-value="${{ inputs.configure-branch }}"'
     - '${{ inputs.config }}'
-
-branding:
-  icon: 'book-open'
-  color: 'orange'
--- a/.github/actions/configure-guides-step/action.yaml
+++ b/.github/actions/configure-guides-step/action.yaml
@@ -5,18 +5,22 @@ inputs:
   config:
     description: 'Configuration file location (i.e. "Documentation/")'
     default: 'Documentation'
+    required: true
 
   configure-branch:
     description: 'guides.xml configuration override (optional): Name of rendered GIT branch (i.e. "main" or "9.5"). Overrides <extension edit-on-github-branch="X">'
     default: ''
+    required: false
 
   configure-project-release:
     description: 'guides.xml configuration override (optional): Project release version (i.e. "draft" or "main"), based on deployment target directory. Overrides <project release="X">'
     default: ''
+    required: false
 
   configure-project-version:
     description: 'guides.xml configuration override (optional): Project version (i.e. "draft", "main", "9.5"). Overrides <project version="X">'
     default: ''
+    required: false
 
 runs:
   using: 'docker'
--- a/.github/actions/configure-guides-step/action.yaml
+++ b/.github/actions/configure-guides-step/action.yaml
@@ -1,0 +1,37 @@
+name: 'configure-guides-step'
+description: 'Configuration for Documentation render tool for typo3 documentation'
+
+inputs:
+  config:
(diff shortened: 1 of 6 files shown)


garvinhicking:
(Addendum: Don't mind the failing "test-self-action" step. It can only succeed once the newly introduced step is merged into `main`.)

garvinhicking:
Another addendum:

Actually I just see that this is currently done in an interesting way via:

```        
run: "sed -i 's/^  image:.*$/  image: \"Dockerfile\"/g' action.yml"
```

in the `workflows/main.yaml` file. I'd need to do the same to the updated action.yml file with the `uses:` attributes. I can do that tomorrow with a fresh mindset, but IMO it shouldn't block merging the PR.

action.yml
```diff
@@ -3,33 +3,51 @@ description: 'Documentation render tool for typo3 documentation'
...
     default: './'
 
   config:
-    description: 'Configuration file location'
+    description: 'Configuration file location (i.e. "Documentation/")'
     default: 'Documentation'
 
   output:
-    description: 'Output directory'
+    description: 'Output directory (i.e. "RenderedDocumentation/Result/project/0.0.0")'
```
brotkrueml, line 20: Should we really have this as example in the description? This nested structure is not necessary anymore, isn't it?

linawolf (approved):
Looks good to me as far as I can tell

action.yml
linawolf: I think that is the structure that is expected for the rest of the deployment process

brotkrueml: Ah, so once introduced, that's the way ;-) But the default value below is another folder name.

garvinhicking: I actually used this default value due to the action in `gh-render-action` https://github.com/TYPO3-Documentation/gh-render-action/blob/main/action.yml#L62: 

```
    - uses: TYPO3-Documentation/render-guides@main
      id: render-guides
      if: steps.enable_guides.outputs.GUIDES == 'true'
      with:
        working-directory: t3docsproject
        config: ./t3docsproject/Documentation/
        output: ./RenderedDocumentation/Result/project/0.0.0
```

So I wanted to make the example relate to how we actually use it.

garvinhicking: (That path is only used during the CI process, that's also why the version number and so forth is fixed)

brotkrueml: Ack.

brotkrueml (approved):

jaapio (commented):
Just a small remark, apart from that, it looks very good.

.github/actions/configure-guides-step/action.yaml
```diff
@@ -0,0 +1,33 @@
+name: 'configure-guides-step'
+description: 'Configuration for Documentation render tool for typo3 documentation'
+
+inputs:
+  config:
+    description: 'Configuration file location (i.e. "Documentation/")'
+    default: 'Documentation'
+
+  configure-branch:
+    description: 'guides.xml configuration override (optional): Name of rendered GIT branch (i.e. "main" or "9.5"). Overrides <extension edit-on-github-branch="X">'
```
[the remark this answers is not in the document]
garvinhicking: Thanks, added!

PR #176, opened 2023-12-13
[DOCS] Add info that at some point we will use `:stable` tagging

garvinhicking:


Documentation/Installation/Index.rst
```diff
@@ -42,6 +42,9 @@ For all available tags, please check the `GitHub packages page`_.
 Once you have pulled the image, you can run the image to render your project's
 documentation.
 
+Note that we will use `:stable` at a later point to always provide stable
```
brotkrueml, line 45: What about `.. note::`` ;-)

Documentation/Migration/Index.rst
```diff
@@ -20,7 +20,8 @@ and optionally also remove the old :file:`Settings.cfg` file.
 With official Docker container
 ==============================
 
-(Work in progress)
+(Work in progress - Note that we will use `:stable` at a later point to always provide stable
```
brotkrueml, line 23: Here too.

README.rst
```diff
@@ -25,7 +25,10 @@ Usage with Docker (via supplied container)
     # On macOS you need to specify the parameter "--user=$(id -u):$(id -g)"
     docker run --rm --pull always -v $(pwd):/project -it ghcr.io/typo3-documentation/render-guides:main --config=Documentation
 
-(see :ref:`_Setup_Docker:Docker containers` for complete documentation)
+(see :ref:`_Setup_Docker:Docker containers` for complete documentation. Note that eventually instead of
```
brotkrueml, line 28: Here too, possibly use `.. seealso::` for the link before.

Documentation/Migration/Index.rst
linawolf: I dont think we need to use a note admonition whereever we use the word "note"...

Documentation/Installation/Index.rst
garvinhicking: I still can't write RST format organically :D Damn, you noticed!

(Will add it now that you mentioned the syntax *g*)

Documentation/Migration/Index.rst
garvinhicking: Please check the rework.

I don't really like that it's at three places now, but that'll be gone one's the notice is gone...

linawolf: I think one note should be enough

brotkrueml: > I dont think we need to use a note admonition whereever we use the word "note"...

Yup, but this should be visible more prominent.

brotkrueml (approved):

PR #178, opened 2023-12-13
[TASK] Use HTML 5 article and section tags

linawolf:


garvinhicking (approved):
(Can't comment much but because it's
blocking progress, go ahead and we fix issues after the fact)

PR #180, opened 2023-12-13
[BUGFIX] Make styles closer to what they were

linawolf:
This is still work in progress but some first fixes. 

Before the change: All links are underlined, fonts not loaded:

image

After change Links only underlined on hover, fonts loaded, all generated scss written into one file

image

I think this will give us a good start to work on the other issues


packages/typo3-docs-theme/assets/sass/_text.scss
```diff
@@ -0,0 +1,7 @@
+/* Bootstrap 5 tweak: do not underline links unless hovered over */
+a:not([class*="btn"]) {
+    text-decoration: none;
+}
+a:not([class*="btn"]):hover {
```
brotkrueml, line 5:
```suggestion
a:not([class*="btn"]):hover,
a:not([class*="btn"]):focus {
```

This way it the underline is also displayed when using a keyboard.

packages/typo3-docs-theme/assets/sass/_webfonts.scss
```diff
@@ -0,0 +1,161 @@
...
+    font-style: italic;
+    font-weight: 400;
+    src: url('../fonts/source-sans-3/source-sans-3-v15-latin_latin-ext-italic.woff2') format('woff2'); /* Chrome 36+, Opera 23+, Firefox 39+, Safari 12+, iOS 10+ */
+}
+/* source-sans-3-500 - latin_latin-ext */
+@font-face {
+    font-display: swap;
+    font-family: 'Source Sans 3';
+    font-style: normal;
+    font-weight: 500;
```
brotkrueml, line 87: So we really need 500, 600, 800 and 900?

400 is normal
700 is bold

packages/typo3-docs-theme/assets/sass/_text.scss
linawolf: I supposed that underlined text in that black focus box do not really improve readability?

image

packages/typo3-docs-theme/assets/sass/_webfonts.scss
linawolf: yeah you are right, 700 and regular with their italics should be enough. My designer in the company always wants all the weights but we dont seem to use them. Changed it

packages/typo3-docs-theme/assets/sass/_text.scss
linawolf: It is also not used currently:
image

brotkrueml: Ah, so we have an outline. Okay then for me, sorry, haven't checked.

brotkrueml (approved):

PR #181, opened 2023-12-13
[TASK] Implement "latest" docker container usage, adjust and document workflows/actions

garvinhicking:
This PR addresses changes in the process of building docker images, and also "baking in" a stable version for CDN asset urls.

The general changes are:

* We only use the GIT `main` branch, like before.
* PRs workflow continues with custom GIT branches, like before. 
* PRs will no longer build docker images or the PHAR. These will be automatically built.
* PRs will still execute the unit and integration test workflows
* Now the docker image will only be built, when we push a version tag to GIT (i.e. `git tag 1.0.1 && git push --tags`).
* A docker-build test-image will also be built upon changes to the `Dockerfile`. That image will not be uploaded.
* PHAR generation only on merge to "main" and on releases

For a description of the full workflow, please see the `Documentation/Developer/Building.rst` with changes included in this PR.

(Please note I had no time to spellcheck/proof-read that or verify that the RST markup works)


linawolf (approved):
Looks good as far as I can tell

.github/actions/configure-guides-step/action.yaml
```diff
@@ -24,7 +24,7 @@ inputs:
 
 runs:
   using: 'docker'
-  image: 'docker://ghcr.io/typo3-documentation/render-guides:main'
+  image: 'docker://ghcr.io/typo3-documentation/render-guides:stable'
```
jaapio, line 27: Think this should be kept on main for now. As you want to have the latest changes for testing?

.github/actions/render-guides-step/action.yaml
```diff
@@ -22,7 +22,7 @@ inputs:
 
 runs:
   using: 'docker'
-  image: 'docker://ghcr.io/typo3-documentation/render-guides:main'
+  image: 'docker://ghcr.io/typo3-documentation/render-guides:stable'
```
jaapio, line 25: Same here

brotkrueml:
Discussed this with @garvinhicking - have open questions, will block it in the meantime.

jaapio (commented):
A strategy should be chosen to change the tag of the docker image before a release. Because you don't want to keep it on latest/main as this would change the executed tag when creating a new release.

garvinhicking:
> A strategy should be chosen to change the tag of the docker image before a release. Because you don't want to keep it on latest/main as this would change the executed tag when creating a new release. 

Yes, Chris had a similar notion, and I understand, the current stable/main workflow is not fleshed out.

I'll try to come up with a variant that works based on tagging only (and always keeps a "main" alias for untagged, plus "latest" for the last released version). So then we could resort to using "latest" for the docker images. 

Then we could reference actions by @main too, if we drop the whole "stable" branch idea.

Or what was your idea, could you elaborate on that more? I'm having tunnelvision and may not see propper alternatives.

The most important thing is that tagged docker images need to have this typo3azureedgeuri constant baked in, that needs to contain the version of the last tag when the docker image was created. "Latest" will always ve overwritten, and old version remain as :major / :major.minor / major.minor.patch because the images remain static once build....

Thanks all!

jaapio:
It might be something to dig into. The github action that is used to build the docker image has multiple ways of tagging. Normally docker images use it the following pattern.

Given we tag V1.1.4

- Latest is always the latest tag. (Semversion)
- V1 == latest version tagged in major version range V1.

- V1.1. == latest version tagged in minor version v1.1

- V1.1.4 === targeting the specific version.

- If we tag v1.1.5 this will result in an update of V1, v1.1 and latest. A new v1.1.5 tag is added.

- If we tag v1.2, the tag V1 and latest are updated. 

GitHub actions use this same pattern. To doing a release would mean a push to branches, v1, v1.1 and tag v.1.1.5

From my experience semversioning for applications is very hard, so I would recommend you to avoid that. I mostly use minor and major versions. If the output and usage is mostly the same it's a minor version. If something drastically changed, it's a major version. 

This becomes harder when you have layered projects. For example, when we do a breaking change in phpDocumentor/guides which requires you to change something in this app, but does not change the way users interact, but it changes the templates? Is that a breaking change for you? 

More simple would be to ignore tags. And just work with a stable branch. A push to this branch would just updated the docker image tag latest and stable to be the state of head in stable.
People should use stable if they want to have an always working version. (Be aware for breaking changes in actions)
Main is your dev version. Being bleeding edge, but not guaranteed to work always.

The latter is more based on the thoughts of trunk based development. This makes your release flow way easier.

garvinhicking:
Thank you @jaapio that helps in thinking.

We do need version numbers for our CDN links (of the typo3-docs-theme subrepo). So we do need to tag releases here, because thag tag is the version for the cdn. We cannot use "stable" for the CDN because when a file is updated in the same directory, it remains cached.

I'll chew on this. 👍

linawolf:
Can we maybe use explicits git tags and whenever we create a new tag a new version is released?

garvinhicking:
Yes, that would be the plan. I've got a few ideas and will write it down tomorrow as a strategy, with the required changes.

brotkrueml (commented):
YOu now use "latest" but refer to "stable" in the PR title and the comments.

garvinhicking:
> You now use "latest" but refer to "stable" in the PR title and the comments.

Odd, I changed that this morning but something didn't save. Well, changed it again in the PR title.

Where else did you see "stable" references? Feel free to push word fixes and so directly to the PR.

PR #184, opened 2023-12-13
[FEATURE] Introduce copy to clipboard for code blocks

brotkrueml:


packages/typo3-docs-theme/resources/template/structure/layout.html.twig
```diff
@@ -162,6 +162,7 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
 <script src="{{ getRelativePath('_resources/js/popper.min.js') }}"></script>
 <script src="{{ getRelativePath('_resources/js/bootstrap.min.js') }}"></script>
 <script src="{{ getRelativePath('_resources/js/theme.min.js') }}"></script>
+<script src="{{ getRelativePath('_resources/js/code-clipboard.min.js') }}"></script>
```
linawolf, line 165: How come you did not include this in the minified theme.min.js?

brotkrueml: The theme is full of jQuery and other stuff and should be cleanup up anyway. I will do it probably later, perhaps only include it when code-blocks are available on the page, don't know yet. But when include it in the theme, this will be done combining via Grunt.

linawolf (approved):

PR #190, opened 2023-12-14
[FEATURE] Use dark blue (tertiary) for link color

linawolf:
The Orange on white text we have currently with links has a very bad contrast:

image

If we tune down the orange to be darker it starts to look like quite a shitty brown. Users have to read a lot of texts with many links in our docs. I would suggest to switch to the tertiary color as a dark blue for link color for now which has good values:

image

The page would then look like this:

image

This greatly improves a11y and readability. If the relaunch of typo3.org offers a better solution for link design we can still adjust to their decisiions later on. We could use the same color for the edit on github button


brotkrueml (commented):
The dark blue color is mostly not distinctable from the text color on my screen, the orange was better recognizable (but less readble). A link should be underlined or has a better color (contrast) to be seen as a link.

linawolf:
I like the idea to underline links as transporting meaning in color is not considered accessible any way

linawolf:
With underlined links we would have this:

image

brotkrueml:
Breadcrumb and TOC can be considered as navigation, maybe not underlined. But for now it is okay, I think. Can be adjustedlater.

brotkrueml (approved):

PR #197, opened 2023-12-14
[Bugfix] Fix colors of phpdomain styles

linawolf:
before:

image

after:

image


brotkrueml (approved):
Fine for me, but we should also consider adding a basic stylesheet to the guides-php-domain package that users can use out of the box.

PR #198, opened 2023-12-14
[TASK] Confval styles

linawolf:
before:

image

after:

image

old style:

image

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/_component_rst.scss
+++ b/packages/typo3-docs-theme/assets/sass/_component_rst.scss
@@ -1,50 +1,6 @@
 .rst-content {
 
     /**
-     * Headlines
-     */
-    h3 {
-        padding-bottom: .5rem;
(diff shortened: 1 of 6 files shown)


packages/typo3-docs-theme/assets/sass/_component_fieldlist.scss
```diff
@@ -25,6 +7,7 @@
         padding-left: 0;
         padding-right: .5em;
         margin-bottom: $spacer * .25;
+        hyphens: manual;
```
brotkrueml, line 10: "manual" is the default value, so no need to define it:
https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens?retiredLocale=de#values

linawolf: removed it

brotkrueml (approved):

PR #208, opened 2023-12-15
[TASK] New styling for version directives

linawolf:
I would like to suggest the following styles for version changed, deprecated etc directives:

image

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/_component_rst.scss
+++ b/packages/typo3-docs-theme/assets/sass/_component_rst.scss
@@ -33,30 +33,7 @@
     //    margin: 0;
     //}
 
-    /**
-     * Version Added, Changed, Deprecated
-     */
-    div {
-        .deprecated,
-        .versionadded,
-        .versionchanged {
-            border-left: 2px solid hsl(26.5, 84.3%, 85%);
-            padding-left: 1em;
-        }
-    }
 
-    /**
-     * Version Modified
-     */
-    .versionmodified {
-        background-color: hsl(26.5, 84.3%, 95%);
-        border-radius: 3px;
-        border: 1px solid hsl(26.5, 84.3%, 85%);
-        color: $code-color-typoscript;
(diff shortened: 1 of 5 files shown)


brotkrueml:
I like the left line and the icons in general (but maybe not an hourglass for deprecation).

But: I don't like the background and border surrounding the text. IMHO the left line is enough to draw attention.

brotkrueml:
Maybe instead of using italic style make it bold?

linawolf:
It is an hourglass where the time has run out, we can use any free fontawesome 6 icon, you have a different sugestion? https://fontawesome.com/search?q=time&o=r&m=free

linawolf:
So how about this?

image

brotkrueml:
Yup, much better. Maybe we leave the hourglass as is, xmark would also be a possibility.

But left-align the icons visually to the text.

How does it look like when you have a note or a warning in that directives? We have them in some places.

garvinhicking:
https://fontawesome.com/icons/skull?f=classic&s=solid

https://fontawesome.com/icons/circle-stop?f=classic&s=regular

https://fontawesome.com/icons/ban?f=classic&s=solid

https://fontawesome.com/icons/road-barrier?f=classic&s=solid

https://fontawesome.com/icons/toilet?f=classic&s=solid

https://fontawesome.com/icons/box-archive?f=classic&s=solid

https://fontawesome.com/icons/calendar-xmark?f=classic&s=solid

garvinhicking:
I like the looks of this!

brotkrueml:
> https://fontawesome.com/icons/ban?f=classic&s=solid

This would be my favourite :) "DON'T USE IT" ;-)

linawolf:
I would suggest if one of these directives is used within an admonition we drop the left border:

image

brotkrueml (approved):
Looks good.

I meant the other way round, but let's get this merged ...

PR #210, opened 2023-12-15
[TASK] Revise TypoScript code highlighting

brotkrueml:
- Remove highlighting of functions
- Remove highlighting of keys in function references
- Add highlighting of EXT: and FILE:EXT: paths


linawolf (approved):
Looks good as far as I can tell

PR #215, opened 2023-12-17
[TASK] Combine custom JavaScript files into one minified file

brotkrueml:
This way, we can use separate source JavaScript files, one for each purpose. They will be combined upon building into one `theme.min.js` file. In a follow-up the "god" theme.js file will be split into separate files.


packages/typo3-docs-theme/resources/template/structure/layout.html.twig
```diff
@@ -160,7 +160,6 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
 <script src="{{ getRelativePath('_resources/js/popper.min.js') }}"></script>
 <script src="{{ getRelativePath('_resources/js/bootstrap.min.js') }}"></script>
 <script src="{{ getRelativePath('_resources/js/theme.min.js') }}"></script>
-<script src="{{ getRelativePath('_resources/js/code-clipboard.min.js') }}"></script>
```
linawolf, line 163: We could also include popper and bootstrap into the minified file, so that we only need to load one file?

linawolf (approved):

packages/typo3-docs-theme/resources/template/structure/layout.html.twig
brotkrueml: Yup, that is another story. But I an not really sure, if we really need popper.

PR #217, opened 2023-12-17
[BUGFIX] Remove unavailable "Source Sans 3" font files

brotkrueml:
We don't provide the removed files in the "fonts" folder, therefore the style definitions are removed.


linawolf (approved):
I thought I removed them. strange

PR #218, opened 2023-12-17
[TASK] Adjust structure of versionchange directives

brotkrueml:
The overall `<section>` groups the whole part. The `<header>` element holds then the heading of the element, which is "Deprecated in version x.y", etc. The text content is then just a paragraph, not a separate article like before.

### DIFF
--- a/packages/typo3-docs-theme/resources/template/body/version-change.html.twig
+++ b/packages/typo3-docs-theme/resources/template/body/version-change.html.twig
@@ -1,16 +1,15 @@
-<div class="versionchange {{ node.type }}">
-    <p class="versionmodified">
+<section class="versionchange {{ node.type }}">
+    <header class="versionmodified">
         <span class="versionicon">
-        {%- if node.type=="deprecated" -%}
-            <i class="fa-solid fa-ban"></i>
-        {%- elseif node.type=="versionadded" -%}
-            <i class="fa-solid fa-rocket"></i>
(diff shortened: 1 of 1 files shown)


packages/typo3-docs-theme/resources/template/body/version-change.html.twig
```diff
@@ -1,16 +1,15 @@
-<div class="versionchange {{ node.type }}">
-    <p class="versionmodified">
+<section class="versionchange {{ node.type }}">
```
linawolf, line 1: should we use "section" here? a section should usually have a heading and for now we only use it for sections in the sense of rst-section so containing one headline and the content.

linawolf: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section?retiredLocale=de "Sections should always have a heading"

brotkrueml: Yup, we have a heading (`<header>`). Another solution would be just use a `<div>` and remove the `<header>`.

linawolf: The header is not a heading. A heading is <h1> - <h6>

linawolf: `h1`- `h6`

PR #219, opened 2023-12-17
[TASK] Refine TypoScript highlighting

brotkrueml:
- Remove number recognition as it has no real value
- Add missing keywords
- Enhance path recognition
- Handle constants (`{$foo.bar}`)
- Distinct between RGB color and comments (`#000` and `#000000`)
- Distinct between SQL quotes and comments ('#fieldname')
- Add tests

### DIFF
--- a/packages/typo3-docs-theme/resources/languages/typoscript.json
+++ b/packages/typo3-docs-theme/resources/languages/typoscript.json
@@ -3,12 +3,37 @@
         "tsconfig"
     ],
     "case_insensitive": false,
-    "keywords": "ADJUST BOX CASE COA COA_INT CONTENT CROP EFFECT ELLIPSE EMBOSS EXTBASEPLUGIN FILES FLUIDTEMPLATE HMENU IMAGE IMG_RESOURCE LOAD_REGISTER OUTLINE PAGE RECORDS RESTORE_REGISTER SCALE SHADOW SVG TEXT TMENU TMENUITEM USER USER_INT WORKAREA",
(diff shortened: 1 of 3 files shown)


linawolf (approved):
While you are at it, I would recommend to make comment detection more strict: 

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/Configuration/TypoScript/Syntax/Comments/Index.html

// and # indicate a comment. Everything until the end of the line will be ignored by the parser. /* indicates a multiline comment start, */ stops it.

When using //, # and /* after an assignment =, this is not considered a comment, but part of the value! Same is true for multiline assignments.

brotkrueml:
Hmm, may be a possibility, but we have than to take care it is non-greedy. But we are talking here about highlighting in the docs. And I can't think of an example, where we use a multi-line comment with an invalid part after the closing comment part.

linawolf:
In TypoScript we do not support the the partly commenting of a line. The whole line is either a comment or not

So for one liner comments you can catch them by `^\n*#`

brotkrueml:
> In TypoScript we do not support the the partly commenting of a line. The whole line is either a comment or not
> 
> So for one liner comments you can catch them by `^\n*#`

This is not true:

    foo < bar // This is a comment
    foo < bar /* This is a valid comment, too */

    foo > # Another valid comment

    foo := addToList(1) # Yes, a comment

    [foo = bar] # Many comment. Much wow.

    foo (
      # This is NOT a comment but part of the value assignment!
      bar = barValue
    ) # This is a comment

brotkrueml:
Okay, don't get the regexp so this one does not match:

    foo = bar // This is NOT a comment but part of the value assignment!

but this one:

    foo // This is a comment

Tried the negative lookbehind assertion but that doesn't work. Also some other possibilities. Will skip.

PR #220, opened 2023-12-17
[TASK] Split up custom JavaScript files

brotkrueml:
The code was moved without modification from the previous `theme.js` file. This makes it easier to get the purpose of a file. They are combined again to `theme.min.js` upon build.

The initialization of `window.T3Docs = {};` was removed as this is not used anymore (as far as I can tell).


brotkrueml:
Next step would be to eliminate jQuery.

linawolf (approved):

PR #223, opened 2023-12-18
[TASK] Revise "menu-expandable.js"

brotkrueml:
- Embed logic into self-executing function so no details are leaked into global space
- Use modern JavaScript constructs like "const" and "forEach"
- Remove usage of jQuery


linawolf (commented):
Where you able to test this without the menus properly working in HTML?

linawolf (approved):
'I guess if nessecary we can make a follow up after subsequent HTMl chagnes

brotkrueml:
> Where you able to test this without the menus properly working in HTML?

Yup, I faked them by tweaking the template to always show "current". Also checked with a page called "TYPO3 Exceptions".

brotkrueml:
> 'I guess if nessecary we can make a follow up after subsequent HTMl chagnes

Yup, of course. The current structure (CSS classes and selectors) is not nice.

Issue #222, opened 2023-12-18
Current state in menu not available


brotkrueml:
When navigating to a menu entry, that menu entry should have the "current" class assigned:

sitemap-level.html.twig:
```html
<li class="toctree-l{{ menuEntry.level }} {%- if menuEntry.isInRootline %} current {% endif %}">
```

menu-level.html.twig may be also affected.

However, it is not:

![Bildschirmfoto vom 2023-12-18 10-17-56](https://github.com/TYPO3-Documentation/render-guides/assets/2566282/99c4b4e3-c377-4aea-a9fc-dd8fae29ce00)

linawolf:
Yes, I am currently waiting for https://github.com/phpDocumentor/guides/pull/759 to be merged

PR #228, opened 2023-12-22
[BUGFIX ]Mark current and active state in Menu

linawolf:
If you open page "Theme Customization" the menu will look like this:

image

We can improve HTML and design in a follow up. 

Resolves https://github.com/TYPO3-Documentation/render-guides/issues/222


packages/typo3-docs-theme/resources/public/css/theme.css
```diff
@@ -1,6 +1,6 @@
 @charset "UTF-8";
 /*!
- * Font Awesome Free 6.5.1 by @fontawesome - https://fontawesome.com
+ * Font Awesome Free 6.4.2 by @fontawesome - https://fontawesome.com
```
brotkrueml, line 3: Please run `npm ci` on the theme folder. In a previous PR I updated the node deps.

brotkrueml (changes requested):

packages/typo3-docs-theme/resources/public/css/theme.css
linawolf: this doesnt change the file. My version of fontawesome is higher then the one before for some reason

linawolf: ok, I hope I worked it out

packages/typo3-docs-theme/resources/template/body/menu/mainmenu/menu-level.html.twig
```diff
@@ -1,19 +1,16 @@
-{%- apply spaceless -%}
-        <li class="toctree-l{{ menuEntry.level }} {%- if menuEntry.isInRootline %} current {% endif %}">
-            <a href="{{ renderLink(menuEntry.url) }}"
-               class="nav-link {% if menuEntry.options.active %} active {% endif %}"
-                {%- if menuEntry.options.active %}  aria-current="page" {% endif -%} >
-                {{ renderNode(menuEntry.value.value) }}
-            </a>
+<li class="toctree-l{{ menuEntry.level }} {%- if menu.currentPath == entry.url %} current {%- endif -%}{% if entry.url in menu.rootlinePaths %} active {%- endif -%}">
```
brotkrueml, line 1: Why use class "current" when no style is attached to it?

linawolf: I think that the menu needs a new and thoughtful HTML structure anyway and thought we can do that in a follow up.

brotkrueml: I don't like loose ends. But when you come with a follow-up in the near time, it's fine.

brotkrueml (approved):

PR #230, opened 2023-12-23
[FEATURE] Display Modal to copy permalinks

linawolf:
If there is a dedicated anchor:

image

If not:

image


brotkrueml (commented):
When displaying the rendered docs locally, there is a "null" before the path:

Bildschirmfoto vom 2023-12-23 15-00-15

brotkrueml:
Maybe an input field is here enough (like for the URL above)?

Bildschirmfoto vom 2023-12-23 15-03-14

brotkrueml:
A link tag does not make sense for local paths:
Bildschirmfoto vom 2023-12-23 15-04-17

brotkrueml:
I would like to see the headers ("Permalink", "reStructuredText (reST)", "HTML") as a header for structuring reasons. Since the header of the modal is a h5, this would be a h6.

brotkrueml:
This not really a link that was copied, maybe "snippet" or similar:
Bildschirmfoto vom 2023-12-23 15-08-46

Also here:
Bildschirmfoto vom 2023-12-23 15-10-00

brotkrueml:
"Create permalink" as heading of the modal feels not right:
Bildschirmfoto vom 2023-12-23 15-11-55

"Create permalink" -> Permalink, rest, HTML ... so maybe something like "Show references"?

Documentation/guides.xml
```diff
@@ -9,6 +9,8 @@
         class="\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension"
         edit-on-github="TYPO3-Documentation/render-guides"
         edit-on-github-branch="main"
+        official-manual-interlink-shortcode="t3docsRenderGuides"
```
brotkrueml, line 12: Hmm, not convinced that this is the way: When I use this, for example, in my EXT:schema (like "ext_schema" as shortcode). Then the user who uses this reference in their extension's docs has to configure it manually.

This may be fine, but we have to hint about this in the modal.

Documentation/guides.xml
```diff
@@ -9,6 +9,8 @@
...
+        official-manual-interlink-predefined="true"
```
brotkrueml, line 13: Why this predefined? What's the consequences?

packages/typo3-docs-theme/assets/js/code-clipboard.js
```diff
@@ -18,4 +18,5 @@
       navigator.clipboard.writeText(codeBlockElement.textContent);
     });
   });
+
```
brotkrueml, line 21:
```suggestion
```

Unrelated to the change.

linawolf:
> so maybe something like "Show references"?

How about "reference this headline"?

brotkrueml:
packages/typo3-docs-theme/assets/sass/components/_modal.scss

is an empty file, so it should be removed?

brotkrueml:
> > so maybe something like "Show references"?
> 
> How about "reference this headline"?

Fine, but we should then also rename the target:

    target="#permalinkModal"

packages/typo3-docs-theme/resources/template/structure/header/additionalMetaData.html.twig
```diff
@@ -0,0 +1,9 @@
+{%- set officialManualShortcode = getSettings('official_manual_interlink_shortcode') -%}
+{%- set isOfficialManualPredefined = getSettings('official_manual_interlink_predefined') -%}
+
+{%- if isOfficialManualPredefined -%}
+    <meta name="t3docs-guides:interlink-predefined-manual" content="true">
```
brotkrueml, line 5: What are these both meta tags about? Where are they used?

packages/typo3-docs-theme/resources/template/structure/layoutParts/permalinkModal.html.twig
```diff
@@ -0,0 +1,52 @@
...
+                    </div>
+                    <p>Copy and freely share the link</p>
+                </div>
+                <div class="mb-3">
+                    <div class="alert alert-warning alert-permalink-rst" role="alert">
+                        {%- set gitHubLink = getEditOnGitHubLink() -%}
+                        This link target has no permanent anchor assigned. {%- if (gitHubLink)  %}
+                        You can make a <a href="{{ getEditOnGitHubLink() }}">pull request on GitHub</a> to suggest an anchor.
+                        {% endif -%}
+                        The link bellow can be used but is prone to change if the page gets moved.
```
brotkrueml, line 30:
```suggestion
                        The link below can be used, but is prone to change if the page gets moved.
```

packages/typo3-docs-theme/assets/js/permalink.js
```diff
@@ -0,0 +1,85 @@
...
+    const item = event.relatedTarget;
+    const section = item.closest('section');
+    const rstAnchor = section ? section.dataset.rstAnchor : null;
+    const headerElement = item.closest('h1, h2, h3, h4, h5, h6');
+    const headerText = headerElement ? headerElement.innerText : '';
+
+    showHideRstAnchorAlert(permalinkModal, rstAnchor);
+
+    const uri = generateUri(section, rstAnchor);
+    const filename = permalinkModal.dataset.currentFilename;
```
brotkrueml, line 77: When the currentFilename is empty (like an Exception is thrown in the filter which is converted to an empty string), then the whole section should be hidden.

linawolf:
> When displaying the rendered docs locally, there is a "null" before the path:

You can test this when you display the document via ddev.

Documentation/guides.xml
linawolf: I will take care of this in a follow up. I first want to investigate on automatic linking to extension manuals. If that proves not to be possible we can still add more info here and reintroduce the flag weather a manual is predefined

packages/typo3-docs-theme/resources/template/structure/header/additionalMetaData.html.twig
linawolf: Not needed anymore, removed

packages/typo3-docs-theme/assets/js/permalink.js
linawolf: done, that will happen in the singlehtml, however js is currently not loaded there, I made an issue: https://github.com/TYPO3-Documentation/render-guides/issues/233

Documentation/guides.xml
linawolf: removed for now, we might not need it

brotkrueml:
> > When displaying the rendered docs locally, there is a "null" before the path:
> 
> You can test this when you display the document via ddev, However I hide the link alltogether now, if it has no window.location.origin

There were some flaws, I fixed them. For me, that part is okay, we can adjust it later if needed.

I also defined the fields as readonly.

Please also consider my additional comments above.

linawolf:
In my eyes I fixed or answered all your comments. which one is open?

brotkrueml:
All the comments not regarding the code but the visual output.

linawolf:
> This not really a link that was copied, maybe "snippet" or similar: Bildschirmfoto vom 2023-12-23 15-08-46
> 
> Also here: Bildschirmfoto vom 2023-12-23 15-10-00

I adjusted the text

linawolf:
> "Create permalink" as heading of the modal feels not right: Bildschirmfoto vom 2023-12-23 15-11-55
> 
> "Create permalink" -> Permalink, rest, HTML ... so maybe something like "Show references"? Also then in the tooltip when hovering.

I changed the headline and strings

linawolf:
> Maybe an input field is here enough (like for the URL above)?
> 
> If we have a long headline an input field can get quite crammed? This is an example with quite a short headline
> 
> Bildschirmfoto vom 2023-12-23 15-03-14

I believe an input field is to small when you have a long headline

linawolf:
> When displaying the rendered docs locally, there is a "null" before the path:
> 
> Bildschirmfoto vom 2023-12-23 15-00-15

If the URL is local it is not displayed at all now

linawolf:
> I would like to see the headers ("Permalink", "reStructuredText (reST)", "HTML") as a header for structuring reasons. Since the header of the modal is a h5, this would be a h6.

They are labels directly bound to the input fields, which in my eyes makes more sense then a headline...

linawolf:
So I tried to give a reply to each of your visual clues. I think we can also make smaller changes in follow ups.

brotkrueml:
> > Maybe an input field is here enough (like for the URL above)?
> > If we have a long headline an input field can get quite crammed? This is an example with quite a short headline
> > Bildschirmfoto vom 2023-12-23 15-03-14
> 
> I believe an input field is to small when you have a long headline

Yup, but you copy it and not read it? And therefore, then also 2 rows are too less maybe. But I am fine with this now.

brotkrueml:
> > I would like to see the headers ("Permalink", "reStructuredText (reST)", "HTML") as a header for structuring reasons. Since the header of the modal is a h5, this would be a h6.
> 
> They are labels directly bound to the input fields, which in my eyes makes more sense then a headline...

This should work?

    <h6><label ...></h6>

But it's fine for now.

brotkrueml (approved):

linawolf:
> Yup, but you copy it and not read it? And therefore, then also 2 rows are too less maybe. But I am fine with this now.

I like the textareas, because you can draw them larger and smaller, while you cannot change the size of an input. And I sometimes like only copying part of the code...

PR #235, opened 2023-12-24
[FEATURE] Automatic interlinks to Core manuals

linawolf:
I will deal with third party extensions in a follow up

Now all common core manuals are predefined. You can call to them without prefix which gives you the preferred version

t3coreapi

If no preferred version is set, stable is used.

You can call also with different prefixes

t3coreapi_dev
t3coreapi_stable
t3coreapi_oldstable
t3coreapi_latest

What versions these go to is defined in Typo3DocsThemeSettings

The prefixes work for any manual that supports versions.

As Interlink inventories are only querried the first time they are used unused inventories, even in a large number should not influence performance notably.

If a version did not exist yet, for example sys_note in version oldstable the normal warning like for a miss configured manual inventory is displayed and the link not expressed.

### DIFF
--- a/Documentation/Developer/InterlinkInventories.rst
+++ b/Documentation/Developer/InterlinkInventories.rst
@@ -1,0 +1,102 @@
+..  include:: /Includes.rst.txt
+
+..  _InterlinkRepositories:
+
+=====================
+Interlink Inventories
+=====================
+
+Sections in other manuals than the current one can be linked during rendering
+by prefixing an anchor link or page link with the name of the manual.
+By using suffixes the version of the manual to be linked can be
+specified:
+
+..  code-block:: rst
+
+    *   :ref:`TYPO3 Explained, preferred version <t3coreapi:start>`
+    *   :ref:`TYPO3 Explained, main version (development) <t3coreapi_dev:start>`
+    *   :ref:`TYPO3 Explained, stable version (for example 12.4) <t3coreapi_stable:start>`
+    *   :ref:`TYPO3 Explained, old stable version (for example 11.5) <t3coreapi_oldstable:start>`
+
+This would output:
+
+*   :ref:`TYPO3 Explained, preferred version <t3coreapi:start>`
+*   :ref:`TYPO3 Explained, main version (development) <t3coreapi_dev:start>`
+*   :ref:`TYPO3 Explained, stable version (for example 12.4) <t3coreapi_stable:start>`
+*   :ref:`TYPO3 Explained, old stable version (for example 11.5) <t3coreapi_oldstable:start>`
+
+The preferred version can be set in the guides.xml to `dev`, `stable`,
+`oldstable` or a specific minor version, for example `8.7`.
+
+..  code-block:: xml
+    :caption: Documentation/guides.xml
+
+    <?xml version="1.0" encoding="UTF-8" ?>
+    <guides
+        xmlns="https://www.phpdoc.org/guides"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
+    >
+        <project title="Render guides"/>
+        <extension
+            class="\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension"
+            typo3-core-preferred="stable"
+        />
+    </guides>
+
+or
+
+..  code-block:: xml
+    :caption: Documentation/guides.xml, excerpt
+
+    <extension
+        class="\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension"
+        typo3-core-preferred="8.7"
+
+It is not necessary anymore to list each of the standard inventories in the
+guides.xml anymore. If desired you can override or redefine standard interlink
+inventories or define new ones:
+
+..  code-block:: xml
+    :caption: Documentation/guides.xml
+
+    <?xml version="1.0" encoding="UTF-8" ?>
+    <guides
+        xmlns="https://www.phpdoc.org/guides"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
+    >
+        <project title="Render guides"/>
+        <extension
+            class="\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension"
+            typo3-core-preferred="stable"
+        />
+        <!-- explicitly link to version 8.7 of TYPO3 Explained -->
+        <inventory id="t3coreapi_v8" url="https://docs.typo3.org/m/typo3/reference-coreapi/8.7/en-us/"/>
+        <!-- ext_sys_note and ext_sys_note_stable should always link to main-->
+        <inventory id="ext_sys_note" url="https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/"/>
+        <inventory id="ext_sys_note_stable" url="https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/"/>
+    </guides>
+
+The following links:
+
+..  code-block:: rst
+
+    *   :ref:`TYPO3 Explained, always version 8.7 <t3coreapi_v8:start>`
+    *   :ref:`Sys note always goes to main <ext_sys_note:start>`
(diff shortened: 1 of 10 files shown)


Documentation/guides.xml
```diff
@@ -9,6 +9,6 @@
         class="\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension"
         edit-on-github="TYPO3-Documentation/render-guides"
         edit-on-github-branch="main"
+        typo3-core-prefered="stable"
```
brotkrueml, line 12: prefered -> preferred

guides.xml
```diff
@@ -15,4 +15,70 @@
     <output-format>html</output-format>
     <output-format>singlepage</output-format>
     <output-format>interlink</output-format>
+    <inventory id="t3coreapi"
```
brotkrueml, line 18: Please sort them alphabetically by id. Also add some breaking and comments between different types of manuals, like

- Manuals
- System extension
- Other

This improves readability a lot.

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
```diff
@@ -4,17 +4,31 @@
 
 class Typo3DocsThemeSettings
 {
+    /**
+     * @var array<string, string>
+     */
+    private static array $defaultSettings = [
```
brotkrueml, line 10: I suggest a backed enum for that.

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
```diff
@@ -4,17 +4,31 @@
...
+        'typo3_core_dev' => "main",
+        'typo3_core_latest' => "12.4",
```
brotkrueml, line 12: Why distinguish between latest and stable?

brotkrueml (changes requested):
Please also add documentation. Also explain how to use the "default settings" (maybe we can find a better word for it, maybe "variants").

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
linawolf: I didn't intend these as a finite list or enum. Just a way to define default values if certain settings are not set in the guides.xml.

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
linawolf: Once 13.0 is released that would be latest but not yet stable?

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
brotkrueml: Maybe, but this list is finite. Or is there a way to hook into that array?

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
brotkrueml: But we don't have 13.0 manuals, we have only main manuals until 13.4 is released. And 13.4 will then become stable.

linawolf: So I removed it

linawolf:
I applied your suggestions. uppon writting the documentation I noted that some things - like overriding inventories - did not work yet, so I had to restructure this some. Also tried to apply enums but not sure if I used them correctly. Please have a look

Documentation/Developer/InterlinkInventories.rst
```diff
@@ -0,0 +1,102 @@
+..  include:: /Includes.rst.txt
+
+..  _InterlinkRepositories:
+
+=====================
+Interlink Inventories
+=====================
+
+Sections in other manuals then the current can be linked during rendering
```
brotkrueml, line 9:
```suggestion
Sections in other manuals than the current one can be linked during rendering
```

Documentation/Developer/InterlinkInventories.rst
```diff
@@ -0,0 +1,102 @@
...
+
+*   :ref:`TYPO3 Explained, always version 8.7 <t3coreapi_v8:start>`
+*   :ref:`Sys note always goes to main <ext_sys_note:start>`
+*   :ref:`Sys note always goes to main <ext_sys_note_stable:start>`
+
+
+Adding a new TYPO3 version or manual
+====================================
+
+When a new TYPO3 Version is released or the long time support changes you
```
brotkrueml, line 97:
```suggestion
When a new TYPO3 version is released or the long-time support changes you
```

I would avoid the "you", maybe ... "the long-time support changes the according TYPo3 versions can be adjusted directly in the theme in the enum" or something. This makes it clear it is nothing for the average user (although, this is about the render-guides, but to make it clear).

Documentation/guides.xml
```diff
@@ -9,6 +9,11 @@
         class="\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension"
         edit-on-github="TYPO3-Documentation/render-guides"
         edit-on-github-branch="main"
+        typo3-core-preferred="stable"
     />
-    <inventory id="h2document" url="https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/" />
+    <!-- explicitly link to version 8.7 of TYPO3 Explained -->
```
brotkrueml, line 14: Maybe add a comment, that they are really used in the chapter "Interlink inventories", so it is clear, what happens here.

packages/typo3-docs-theme/src/Inventory/DefaultInventories.php
```diff
@@ -0,0 +1,49 @@
...
+    case t3ts45 = 'https://docs.typo3.org/m/typo3/tutorial-typoscript-in-45-minutes/{typo3_version}/en-us/';
+
+    // Team Guides, they are commonly not versioned
+    case h2document = 'https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/';
+    case t3content = 'https://docs.typo3.org/m/typo3/guide-contentandmarketing/main/en-us/';
+    case t3contribute = 'https://docs.typo3.org/m/typo3/guide-contributionworkflow/main/en-us/';
+
+    // System Extensions
+    case ext_adminpanel = 'https://docs.typo3.org/c/typo3/cms-adminpanel/{typo3_version}/en-us/';
+    case ext_core = 'https://docs.typo3.org/c/typo3/cms-core/{typo3_version}/en-us/';
```
brotkrueml, line 29:
```suggestion
    case ext_core = 'https://docs.typo3.org/c/typo3/cms-core/main/en-us/';
```

Changelog is always "main".

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
```diff
@@ -8,13 +8,14 @@ class Typo3DocsThemeSettings
      * @param array<string, string> $settings
      */
     public function __construct(
-        private readonly array $settings
+        private array $settings
```
brotkrueml, line 11: Why remove the "readonly"? Nothing else has changed in this class?

packages/typo3-docs-theme/src/Inventory/DefaultInventories.php
linawolf: true - also as the changelog is really something else then just any system extension manual (and also handled differently with the version) I moved it to its own section.

Documentation/guides.xml
linawolf: done

Documentation/Developer/InterlinkInventories.rst
```diff
@@ -0,0 +1,102 @@
...
+*   :ref:`Sys note always goes to main <ext_sys_note:start>`
+*   :ref:`Sys note always goes to main <ext_sys_note_stable:start>`
+
+
+Adding a new TYPO3 version or manual
+====================================
+
+In the event of a change in long-term support, adjustments to the corresponding
+TYPO3 versions can be made directly in the theme within the enum
+:php:`\T3Docs\Typo3DocsTheme\Inventory\Typo3VersionMapping`.
```
brotkrueml, line 99: The backslashes must be escaped now, which is be a breaking change to the Sphinx-based rendering. Currently it look like:

Bildschirmfoto vom 2023-12-26 15-02-41

Is this known? I assume, this is a bug?

linawolf: I will make an issue and fix it in the renderer. I didnt notice yet. We don't have to change it here as the renderer fix will fix that

brotkrueml (approved):

PR #236, opened 2023-12-24
[CI] Let dependabot create PRs for updating dependencies

brotkrueml:
Composer and npm updates can be created automatically using dependabot. Updates are checked on a daily basis at 16:00. For each pull request created by dependabot the labels "dependencies", and "composer" (for a Composer-related update) and "npm" for a (npm-related update) are added.

Development branched are also considered for Composer-related updates for the time being.

Updates of requirements can also be done manually at any time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "composer"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


brotkrueml:
As far as I know, the monorepo structure should be considered. Let's see ...

linawolf:
Great idea, thanks! I also considered doing something like that. I would opt for trying it out and making adjustments if needed.

the phpdocumentor/guides are doing something similar, we could have a look how they handle dependencies in a monorepo

linawolf (approved):

garvinhicking:
Late to the party, great idea!

PR #248, opened 2023-12-27
[TASK] Move and rename scss files

linawolf:
Where order does not matter, import them alphabetically.

No changes to the files themselves were made.


packages/typo3-docs-theme/assets/sass/fonts/_webfonts.scss
```diff
@@ -4,31 +4,31 @@
     font-family: 'Source Code Pro';
     font-style: normal;
     font-weight: 400;
-    src: url('../fonts/source-code-pro/source-code-pro-v23-latin_latin-ext-regular.woff2') format('woff2'); /* Chrome 36+, Opera 23+, Firefox 39+, Safari 12+, iOS 10+ */
+    src: url('../../fonts/source-code-pro/source-code-pro-v23-latin_latin-ext-regular.woff2') format('woff2'); /* Chrome 36+, Opera 23+, Firefox 39+, Safari 12+, iOS 10+ */
```
brotkrueml, line 7: I don't think this works... the path is now only relative to the scss folder, but not to the generated theme.css - which is wrong.

linawolf: Good catch! I fixed it and replaced the fonts path with a variable for easier adjustment

brotkrueml (approved):

PR #250, opened 2023-12-27
[CI] Approve and merge pull requests by dependabot automatically

brotkrueml:
Only pull request where the tests are passing can be approved and merged due to the setting in the repo.

### DIFF
--- a/.github/workflows/pr-auto-approve.yaml
+++ b/.github/workflows/pr-auto-approve.yaml
@@ -1,0 +1,29 @@
+# https://docs.github.com/en/github-ae@latest/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions
+name: Auto-approve
+
+on:
+  pull_request:
+
+permissions:
+  pull-requests: write
+
+jobs:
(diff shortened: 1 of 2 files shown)


brotkrueml:
Code taken from EXT:tea.

linawolf (approved):

garvinhicking (approved):
Looking good, let's try it!

Btw, How would you feel on doing the dependabots weekly instead of daily? Might make bundled review easier, instead of maybe cluttering up?

brotkrueml:
> Looking good, let's try it!
> 
> Btw, How would you feel on doing the dependabots weekly instead of daily? Might make bundled review easier, instead of maybe cluttering up?

@garvinhicking 
- AFAIK every update is a separate PR.
- It is approved and merged automatically now with the PR, so no need to do something manually (except tests are failing because of the update).
- We can change it later to weekly when the most work is done. But now it is easier to run in daily instead of doing manual PRs because we need a specific library to be updated.
- And: Daily or weekly is not in the scope of this PR ;-)

PR #251, opened 2023-12-27
[TASK] Move processing logic out of MigrateSettingsCommand

brotkrueml:
The command can be seen as a controller, it should not hold logic outside controlling. Therefore, the processing is moved into a dedicated class.

Now we can also test the command class easily.


linawolf (approved):

garvinhicking (approved):
Thanks, this looks like solid refactoring to me, great job. Let's
Go :-)

PR #252, opened 2023-12-28
[TASK] Refactor Menu files

linawolf:
This is  a preparation for the improvement of the menu. Up unto here neither display nor functionality are changed


packages/typo3-docs-theme/assets/js/menu-expandable.js
```diff
@@ -41,7 +41,7 @@
 
 
 document.addEventListener('DOMContentLoaded', () => {
-  const currentFirstLevelEntries = document.querySelectorAll('li.toctree-l1.active');
+  const currentFirstLevelEntries = document.querySelectorAll('.menu-level-1 > li.active');
```
brotkrueml, line 44: Do we need the "li" here in the selector? Querying the tag and the class looks too much?

packages/typo3-docs-theme/resources/template/structure/document.html.twig
```diff
@@ -7,12 +7,10 @@
     {{ parent() }}
 {% endblock %}
 
-{% block navbar %}
-    {% include "structure/navigation/navbar.html.twig" %}
-{% endblock %}
-
 {% block menu_doc %}
-    {% include "structure/navigation/menu_doc.html.twig" %}
+    <div aria-label="Main navigation" class="main_menu" role="navigation">
```
brotkrueml, line 11: Instead of `<div>` use `<nav>`? Then the role would be superfluous.

linawolf: Yes I am still working on these kind of changes, but I try to keep the changes small per PR so they are reviewable There is already a nav tag but it seems to be in the wrong position

linawolf: In this case I moved the block without changing it from one template into another

packages/typo3-docs-theme/assets/js/menu-expandable.js
linawolf: I cannot see any disadvantage in being specific? We querried for li before as well. And also I believe this part of the JS will soon be gone as we are able now to remove the unwanted levels of menus from the HTML output instead of hiding them by JS

brotkrueml (approved):

PR #255, opened 2023-12-29
[TASK] Add tests for the new improved menu

linawolf:
* Menu depth can now be limited by setting `max-menu-depth="1"`
* Main menu is calculated once during compiling instead of for each page during rendering
* `:maxdepth: 1` on the toctree does not limit the main menu anymore (just like in Sphinx)


brotkrueml:
So, max-menu-depth is a new guides attribute? I am missing the context in the attribute or is this setting considered for every menu (like also tocs for a page)?

linawolf:
Yes it is a new setting in the guides, feel free to propose a better name there.

It limits the global menu stored in the project node.

brotkrueml:
Why do we have then tests for it? Shouldn't they be in guides?

linawolf:
They are but as these functionalities are important to us I want to make sure, we don't break them. I didnt add all tests but just two cases. They are test-cases that would not have worked until now

brotkrueml:
Seems, you have little confidence in the guides tests ;-)

brotkrueml (approved):

linawolf:
I do have confidence in the guides tests. But everything the guides do is overridable in multiple ways. So we might break something that works in the guides.

PR #266, opened 2023-12-30
[TASK] Remove popper

brotkrueml:
Popper is required by Bootstrap when using popovers. However, we do not use that feature, so we can safely remove this library.


linawolf:
I was considering to use popovers. For example A popover could improve readability for FQN strings. I was thinking about only displaying the classname in within the text, then supply the FQN and if the PHP class is defined as php domain some additional information.

brotkrueml:
> I was considering to use popovers. For example A popover could improve readability for FQN strings. I was thinking about only displaying the classname in within the text, then supply the FQN and if the PHP class is defined as php domain some additional information.

Hmm, I am currently not convinced that is the way for FQNs. However, when necessary you can add it then again as part of the patch IMHO.

brotkrueml:
I marked it as a draft for now.

garvinhicking:
Since we now use it for display of popovers for the inline code tools, I'll close this here

PR #278, opened 2024-01-06
[FEATURE] Automatic Interlinks for extensions

linawolf:
depends on #277

### DIFF
--- a/Documentation/Developer/InterlinkInventories.rst
+++ b/Documentation/Developer/InterlinkInventories.rst
@@ -8,25 +8,38 @@ Interlink Inventories
 
 Sections in other manuals than the current one can be linked during rendering
 by prefixing an anchor link or page link with the name of the manual.
+
+Interlinks to official manuals
+==============================
+
 By using suffixes the version of the manual to be linked can be
 specified:
 
 ..  code-block:: rst
 
     *   :ref:`TYPO3 Explained, preferred version <t3coreapi:start>`
-    *   :ref:`TYPO3 Explained, main version (development) <t3coreapi_dev:start>`
-    *   :ref:`TYPO3 Explained, stable version (for example 12.4) <t3coreapi_stable:start>`
-    *   :ref:`TYPO3 Explained, old stable version (for example 11.5) <t3coreapi_oldstable:start>`
+    *   :ref:`TYPO3 Explained, main version (development) <t3coreapi/dev:start>`
+    *   :ref:`TYPO3 Explained, stable version (for example 12.4) <t3coreapi/stable:start>`
+    *   :ref:`TYPO3 Explained, old stable version (for example 11.5) <t3coreapi/oldstable:start>`
+    *   :ref:`TYPO3 Explained 12 LTS <t3coreapi/v12:start>`
+    *   :ref:`TYPO3 Explained 12.4 <t3coreapi/12.4:start>`
 
 This would output:
 
 *   :ref:`TYPO3 Explained, preferred version <t3coreapi:start>`
-*   :ref:`TYPO3 Explained, main version (development) <t3coreapi_dev:start>`
-*   :ref:`TYPO3 Explained, stable version (for example 12.4) <t3coreapi_stable:start>`
-*   :ref:`TYPO3 Explained, old stable version (for example 11.5) <t3coreapi_oldstable:start>`
+*   :ref:`TYPO3 Explained, main version (development) <t3coreapi/dev:start>`
+*   :ref:`TYPO3 Explained, stable version (for example 12.4) <t3coreapi/stable:start>`
+*   :ref:`TYPO3 Explained, old stable version (for example 11.5) <t3coreapi/oldstable:start>`
+*   :ref:`TYPO3 Explained 12 LTS <t3coreapi/v12:start>`
+*   :ref:`TYPO3 Explained 12.4 <t3coreapi/12.4:start>`
+
+..  _typo3-version:
+
+Setting the preferred TYPO3 version
+-----------------------------------
 
 The preferred version can be set in the guides.xml to `dev`, `stable`,
-`oldstable` or a specific minor version, for example `8.7`.
+`oldstable` or a specific minor version, for example `8.7` or `v8`.
 
 ..  code-block:: xml
     :caption: Documentation/guides.xml
@@ -54,7 +67,7 @@ or
         typo3-core-preferred="8.7"
 
 It is not necessary anymore to list each of the standard inventories in the
-guides.xml anymore. If desired you can override or redefine standard interlink
+:file:`guides.xml` anymore. If desired you can override or redefine standard interlink
 inventories or define new ones:
 
 ..  code-block:: xml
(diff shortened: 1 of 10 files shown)


Documentation/Developer/InterlinkInventories.rst
```diff
@@ -92,11 +96,61 @@ The following links:
...
 
 The default manuals to be supported can be managed in enum
 :php:`\T3Docs\Typo3DocsTheme\Inventory\DefaultInventories`.
+
+
+Interlinks to TYPO3 extensions
+==============================
+
+You can link to the main version of the manual of a TYPO3 extension if
+that extension's manual has been rendered on our server.
```
brotkrueml, line 113:
```suggestion
that extension's manual has been rendered on docs.typo3.org.
```

Documentation/Developer/InterlinkInventories.rst
```diff
@@ -92,11 +96,61 @@ The following links:
...
+
+To create an interlink to a third-party extension prefix the extensions
```
brotkrueml, line 115:
```suggestion
To create an interlink to a third-party extension prefix the extension's
```

Documentation/Developer/InterlinkInventories.rst
```diff
@@ -92,11 +96,61 @@ The following links:
...
+composer name with `ext/` and use it as interlink domain:
```
brotkrueml, line 116:
```suggestion
Composer name with `ext/` and use it as interlink domain:
```

Documentation/Developer/InterlinkInventories.rst
```diff
@@ -92,11 +96,61 @@ The following links:
...
+    *   :doc:`News <ext/georgringer/news:Index>`
+    *   :ref:`External Imports <ext/cobweb/external_import:start>`
+
+This will be rendered as:
+
+*   :doc:`News <ext/georgringer/news:Index>`
+*   :ref:`External Imports <ext/cobweb/external_import:start>`
+
+
+If an extension author need to link to a specific version of an extensions manual,
```
brotkrueml, line 129:
```suggestion
If an extension author needs to link to a specific version of an extension's manual,
```

Documentation/Developer/InterlinkInventories.rst
```diff
@@ -92,11 +96,61 @@ The following links:
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
+    >
+        <project title="Render guides"/>
+        <extension
+            class="\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension"
+            typo3-core-preferred="stable"
+        />
+        <!-- explicitly link to version 11.3 of the news manual -->
+        <inventory id="ext/georgringer-news-11.3" url="https://docs.typo3.org/p/georgringer/news/11.3/en-us/"/>
```
brotkrueml, line 147: Hmm, so sometimes you use `ext/` and sometimes `ext_` as id - so there is no rule?

Above, there is a path (I assume) on the server for the main branch, here we should stick to `ext_` like suggested today.

Documentation/guides.xml
```diff
@@ -18,4 +18,6 @@
     <!-- ext_sys_note and ext_sys_note_stable should always link to main-->
     <inventory id="ext_sys_note" url="https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/"/>
     <inventory id="ext_sys_note_stable" url="https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/"/>
+    <!-- explicitly link to version 11.3 of the news manual -->
+    <inventory id="ext/georgringer/news-11.3" url="https://docs.typo3.org/p/georgringer/news/11.3/en-us/"/>
```
brotkrueml, line 22: See above, here we use `ext/` as prefix, for sys_note we use `ext_`. We should not confuse users.

packages/typo3-docs-theme/src/Inventory/Typo3InventoryRepository.php
```diff
@@ -2,22 +2,37 @@
...
 use T3Docs\Typo3DocsTheme\Settings\Typo3DocsThemeSettings;
 
 final class Typo3InventoryRepository implements InventoryRepository
 {
+    /**
+     * @see https://regex101.com/r/EXCPkt/1
+     *
+     * https://getcomposer.org/doc/04-schema.md#name
+     */
+    private const EXTENSION_INTERLINK_REGEX = '/^ext\/([^\/]*)\/(.*)$/';
```
brotkrueml, line 22:
```suggestion
    private const EXTENSION_INTERLINK_REGEX = '/^ext\/([^\/]+)\/(.+)$/';
```

To avoid `ext//` which would be possible before.

Documentation/guides.xml
linawolf: for sys_note we also don't use the composer name. I would suggest to phase out the old keys, deprecate them after a while and then switch to composer names for each and every extension, including core extensions?

linawolf: See above

Documentation/Developer/InterlinkInventories.rst
linawolf: ext_sys_notes is not a composer name. I would suggest to phase out the old keys and deprecate them a while after our switch. we can then use composer names for each and every extension, including core extensions

Documentation/guides.xml
brotkrueml: As it is just an id, which references an URL and can be any string, I don't really care about that. But it should be consistent throughout the documentation (not only then guides.xml also the docs). There is no need for deprecation, as it has no impact for the user, If you name it "a" or "b/a". It is just a suggestion or convention.

Documentation/Developer/InterlinkInventories.rst
brotkrueml: But "ext/typo3/cms-sys-note" is.

linawolf: So what exactly do you suggest? Please let us discuss in only one place this gets confusing

Documentation/guides.xml
linawolf: see above

Documentation/Developer/InterlinkInventories.rst
brotkrueml: I like to use the same structure, not a mixture of `ext_`and `ext/`. So, when we go with `ext/` the sys_note should also be adjusted. Or is there a rule, when to use the one or the other I don't get?

brotkrueml: But, anyway, I don't care anymore about that. Do it, when you think that is the way.

brotkrueml (dismissed):

Documentation/Developer/InterlinkInventories.rst
linawolf: @brotkrueml I gave it another try to make it more logic. The ext prefix is not really needed as we didn't have slashes in the interlink domain in sphinx. So using the composer name as interlink domain now works for both third party extensions and system extensions. The legacy names ext_adminpanel etc are kept for now, we can deprecate them later. Also I automatically made the versions linkable. In almost all cases it should not be necessary anymore to define interlinks in the guides.

brotkrueml: I do not getting warm with the syntax. Once it is `t3coreapi/stable`, then it is `typo3/cms-sys-note/stable`. The second one is derived from the Composer package, the first one not. Personally, I would like to have at least another separator for the version, like: `t3coreapi@stable`  and `typo3/cms-sys-note@stable`. Sadly, we can't use the colon (like Composer) as this is already in use for a rst reference.

However, when you think this is the way, go ahead, I won't block it.

brotkrueml (approved):

Documentation/Developer/InterlinkInventories.rst
linawolf: I can investigate if we can use the colon syntax. We are overriding the interlink detection so if we get the regex worked out we can do anything we want to

brotkrueml: After thinking again, colon is more confusing. Think of: `typo3/cms-sys-note:stable:the-ref` That's not worth the work.

brotkrueml: For me it is okay for now. Let's get it in, then it can be adjusted if necessary.

PR #281, opened 2024-01-10
[FEATURE] In Migration drop predefined inventories

linawolf:
and determine preferred TYPO3 version

### DIFF
--- a/packages/typo3-guides-cli/composer.json
+++ b/packages/typo3-guides-cli/composer.json
@@ -11,7 +11,8 @@
     },
     "require": {
         "ext-dom": "*",
-        "ext-libxml": "*"
+        "ext-libxml": "*",
+        "t3docs/typo3-version-handling": "self.version"
     },
     "bin": [
         "bin/typo3-guides"
--- a/packages/typo3-guides-cli/src/Migration/SettingsMigrator.php
+++ b/packages/typo3-guides-cli/src/Migration/SettingsMigrator.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace T3Docs\GuidesCli\Migration;
 
 use T3Docs\GuidesCli\Migration\Dto\MigrationResult;
+use T3Docs\VersionHandling\DefaultInventories;
+use T3Docs\VersionHandling\Typo3VersionMapping;
 
 class SettingsMigrator
 {
@@ -18,6 +20,7 @@ class SettingsMigrator
      */
     private array $unmigratedSettings = [];
     private int $convertedSettings = 0;
+    private string $detectedVersion = 'stable';
 
     /**
      * Return the XML document, the number of converted settings and
@@ -35,9 +38,18 @@ class SettingsMigrator
         $this->xmlDocument->formatOutput = true;
 
         $guides = $this->createRootElement();
-        $this->createExtensionSection($guides);
-        $this->createProjectSection($guides);
-        $this->createInventorySection($guides);
+        $extension = $this->createExtensionSection();
+        $project = $this->createProjectSection();
+        $inventories = $this->createInventorySection();
+        $extension->setAttribute('typo3-core-preferred', $this->detectedVersion);
+
+        $guides->append($extension);
+        if ($project !== null) {
+            $guides->append($project);
+        }
+        foreach ($inventories as $inventory) {
+            $guides->append($inventory);
+        }
 
         $messages = $this->collectUnmigratedLegacySettings();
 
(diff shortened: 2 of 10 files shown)


packages/typo3-docs-theme/src/Inventory/Typo3VersionMapping.php
```diff
@@ -37,4 +37,16 @@ public static function getDefault(): Typo3VersionMapping
     {
         return Typo3VersionMapping::Stable;
     }
+
+    /**
+     * @return list<string>
+     */
+    public static function getAllVersions(): array
+    {
+        $allVersions = [];
```
brotkrueml, line 46: Maybe easier with array_map:

$allVersions = array_map(static fn(string $enumValue) => $enumValue->getVersion(), self::cases());

packages/typo3-guides-cli/composer.json
```diff
@@ -11,7 +11,8 @@
     },
     "require": {
         "ext-dom": "*",
-        "ext-libxml": "*"
+        "ext-libxml": "*",
+        "t3docs/typo3-docs-theme": "dev-main"
```
brotkrueml, line 15: The cli package should not depend on a theme package (used mainly for frontend). Maybe we can move the Typo3VersionMapping enum somewhere else?

We run in problems for releases with the constraint? Maybe we can use "self.version" instead of "dev-main"?

packages/typo3-guides-cli/src/Migration/SettingsMigrator.php
```diff
@@ -35,9 +38,18 @@ public function migrate(array $legacySettings): MigrationResult
...
-        $this->createExtensionSection($guides);
-        $this->createProjectSection($guides);
-        $this->createInventorySection($guides);
+        $extension = $this->createExtensionSection();
+        $project = $this->createProjectSection();
+        $inventories = $this->createInventorySection();
+        $extension->setAttribute('typo3-core-preferred', $this->detectedVersion);
+
+        $guides->append($extension);
+        if ($project != null) {
```
brotkrueml, line 47:
```suggestion
        if ($project !== null) {
```

packages/typo3-guides-cli/src/Migration/SettingsMigrator.php
```diff
@@ -18,6 +20,7 @@ class SettingsMigrator
      */
     private array $unmigratedSettings = [];
     private int $convertedSettings = 0;
+    private string $detectedVersion = 'stable';
```
brotkrueml, line 23: better would be a constant? The value isn't changed?

packages/typo3-guides-cli/src/Migration/SettingsMigrator.php
```diff
@@ -101,28 +113,49 @@ private function createProjectSection(\DOMElement $parentNode): void
...
             $this->convertedSettings++;
 
+            if ($defaultInventory = DefaultInventories::tryFrom($id)) {
+                $defaultUrl = $defaultInventory->getUrl();
+                if ($url === $defaultUrl) {
+                    continue;
+                } else {
+                    $allowedVersions = Typo3VersionMapping::getAllVersions();
+                    $urlDifference = $this->getStringDifference($url, $defaultUrl);
+                    if (in_array($urlDifference, $allowedVersions, true)) {
```
brotkrueml, line 142: Perhaps use Typo3VersionMapping::tryFrom(), this returns the case or null. So no need to call getAllversions() and check for existence of a value.

brotkrueml (changes requested):

packages/typo3-guides-cli/src/Migration/SettingsMigrator.php
linawolf: The allowed versions are the result of $versionType->getVersion(), you cannot get them by calling tryFrom

linawolf:
@brotkrueml I moved the version and inventory enums into their own package. What do oyu think?

brotkrueml (approved):
Code review looks good. Better then before, we'll see where this leads to ;-)

I am not sure, if tests are helping for the enum methods. At least they protect against unintended changes - but have to be adjusted with a new LTS (which would be fine). What do you think?

brotkrueml:
Ah, one thing: T3Docs\VersionHandling as namespace is a little bit vague. Maybe T3Docs\GuidesVersionHandling? Not sure.

linawolf:
I think it should be fine. The versions are not necessarly bound to the guides?

PR #282, opened 2024-01-11
[FEATURE] Add hot-reload preview of ReST file changes

garvinhicking:
A small "proof of concept" that currently only works with ddev.

It allows you to start a small gulp watch task, so that any changes performed on ReST files automatically trigger rendering and reloads your browser (at your current scroll position).

You can use this to have your IDE and a browser next to each other and preview your edits quickly (not realtime, but for me in ~3 seconds after saving).

This could need optimization to only render a single input file instead of everything. And it would be cool to make it work independently of ddev (should be possible, it just needs the configured port, I already added a "localhost" fallback when a ddev variable does not exist).

This would also have potential to be added as a composer package (then then rendering would be done with the docker container, not locally - that might add half a second or so to the run), so extension developers could use it easily on their repositories. For now, this only works on Documentation of this exact project.

### DIFF
--- a/.ddev/docker-compose.rendersync.yaml
+++ b/.ddev/docker-compose.rendersync.yaml
@@ -1,0 +1,10 @@
+#ddev-generated
+# Override the web container's standard HTTP_EXPOSE and HTTPS_EXPOSE
+# This is to expose the browsersync port.
+services:
+  web:
+    expose:
+      - '3000'
+    environment:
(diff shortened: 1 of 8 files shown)


linawolf (approved):
Quite cool!

garvinhicking:
I think this is not the final route to go.

I will extract this as a standalone composer package that can be dev-required, so it can be used for any other repository too. I will also asd docs on how to achieve it with inotify&phpstorm as an alternative.

Then I can add documentation in this repository on how to use the extra package and point to its documentation.

Hope that will help at least one more person than myself 🤓

garvinhicking:
(Please let the feature branch exist until I could move it away)

PR #284, opened 2024-01-11
[TASK] Clean up Admonition templates

linawolf:
It is not really necessary to have separate templates for each admonition type.

Also fix Admonitions that are in one line. They had no content until now

### DIFF
--- a/packages/typo3-docs-theme/resources/template/body/admonition.html.twig
+++ b/packages/typo3-docs-theme/resources/template/body/admonition.html.twig
@@ -1,15 +1,28 @@
 {% if name == 'important' %}
-    {% include "body/admonitions/important.html.twig" %}
+    {% set titleOfSeverity='Important' %}
 {% elseif name == 'attention' %}
-    {% include "body/admonitions/attention.html.twig" %}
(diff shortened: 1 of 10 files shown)


packages/typo3-docs-theme/resources/template/body/admonition.html.twig
```diff
@@ -1,15 +1,28 @@
...
+    {% set titleOfSeverity='Caution' %}
+{% elseif name == 'hint' %}
+    {% set titleOfSeverity='Hint' %}
+{% elseif name == 'tip' %}
+    {% set titleOfSeverity='Tip' %}
 {% elseif name == 'seealso' %}
-    {% include "body/admonitions/seealso.html.twig" %}
-{% else %}
-    {% include "body/admonitions/default.html.twig" %}
+    {% set titleOfSeverity='Seealso' %}
```
brotkrueml, line 16: Maybe we should write "See also" as "Seealso" is no English word.

linawolf: already saw and changed it, forgot to push ;)

brotkrueml (approved):

PR #285, opened 2024-01-13
[BUGFIX] Deactivate PlantUmlServer calls for tests

linawolf:
When the docker container is used in testing pipelines
to test if documentation renders without warnings,
the plantuml server sometimes makes the tests fail by not being available.

This PR disables the plantuml server calls in testing-mode
that is when option --fail-on-log is set or in the integration tests

### DIFF
--- a/Documentation/Index.rst
+++ b/Documentation/Index.rst
@@ -45,3 +45,32 @@ to extend the tool.
     Developer/Index
     Migration/Index
     KnownProblems/Index
+
+..  uml::
+
+    class Foo1 {
+        You can use
+        several lines
+        ..
+        as you want
+        and group
+        ==
+        things together.
+        __
+        You can have as many groups
+        as you want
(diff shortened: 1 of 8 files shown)


brotkrueml:
So, to be clear: We disable a server which is needed to render UMLs, so we get no feedback, if the UML is correct?

linawolf:
The only feedback the plantum server is giving us here is a status code if the server is down. So testing the UML in the sense of catching guides warnings is not possible at the moment. However when the plantuml server is down you will receive a non 200 status code and the tests will fail.

brotkrueml:
Oh okay, so to render the plantuml thing, an external server is needed? Wow.

brotkrueml (approved):

linawolf:
The other possibility would be to supply your own plantuml service, that is also possible, if you want to investigate

Issue #289, opened 2024-01-15
Styling mistake for the Image subtitle


FlorianT03:
At the bottom, below the bottom image, the gap between the image and the "4.Save the plugin." is too large.

Compare:
https://docs.typo3.org/p/georgringer/news/draft/en-us/QuickStart/CreateRecords.html
https://docs.typo3.org/p/georgringer/news/main/en-us/QuickStart/CreateRecords.html

linawolf:
I think it looks better with the larger gap in most cases.

brotkrueml:
The image has a large offset (i added a border):

![image](https://github.com/TYPO3-Documentation/render-guides/assets/2566282/0eb21065-4211-4726-9b15-aac7baa544e9)

So, nothing we can do about it.

brotkrueml:
But: The class is not recognised: The "with-shadow" is not considered. This may be a bug.

linawolf:
yes, looking into the shadow

PR #296, opened 2024-01-15
[BUGFIX] Restore image shadows

linawolf:
This restores the shadow, usefull for images who are white:

image

the shadow is applied if the image was defined with class "with-shadow":

```
..  image:: some-image.png
    :class: with-shadow
```

Additionally restored styles for figures with and without shadows:

image

```

..  figure:: typo3-logo.svg
    :class: with-shadow

    Some Caption



..  figure:: typo3-logo.svg

    Some Caption
```

resolves https://github.com/TYPO3-Documentation/render-guides/issues/289

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/components/_images.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_images.scss
@@ -1,0 +1,17 @@
+img.with-shadow, figure.with-shadow {
+    @extend .shadow;
+    @extend .p-3;
+    @extend .mb-4;
+}
+
+figure {
+    @extend .figure;
+    figcaption {
+        @extend .pt-2;
+        @extend .figure-caption;
+
+        p:last-child {
+            margin-bottom: 0;
+        }
+    }
+}
--- a/packages/typo3-docs-theme/resources/public/css/theme.css
+++ b/packages/typo3-docs-theme/resources/public/css/theme.css
@@ -11156,7 +11156,7 @@ progress {
   height: auto;
 }
 
-.figure {
+.figure, figure {
   display: inline-block;
 }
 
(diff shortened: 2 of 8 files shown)


packages/typo3-docs-theme/assets/sass/components/_images.scss
```diff
@@ -0,0 +1,17 @@
+img.with-shadow, figure.with-shadow {
```
brotkrueml, line 1: Can we just use `.with-shadow` - with the tag is looks over-specified.

packages/typo3-docs-theme/resources/public/css/theme.css
```diff
@@ -18123,7 +18123,7 @@ textarea.form-control-lg {
   display: none !important;
 }
 
-.shadow {
+.shadow, img.with-shadow, figure.with-shadow {
```
brotkrueml, line 18126: here too

packages/typo3-docs-theme/resources/public/css/theme.css
```diff
@@ -11156,7 +11156,7 @@ progress {
   height: auto;
 }
 
-.figure {
+.figure, figure {
```
brotkrueml, line 11159: Oh boy, is this ugly. But okay for now, I think.

packages/typo3-docs-theme/resources/public/css/theme.css
linawolf: This css is automatically generated, it does not make sense to review it unless you are tinkering with the tools who generate it?

packages/typo3-docs-theme/resources/public/css/theme.css
linawolf: This css is automatically generated...

packages/typo3-docs-theme/assets/sass/components/_images.scss
brotkrueml: Ack

brotkrueml (approved):

PR #308, opened 2024-01-21
[TASK] Icons for Inline-code languages

linawolf:
After:

image

Before:

image

(I am open for a better icons here ;-) )

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/_variables.scss
+++ b/packages/typo3-docs-theme/assets/sass/_variables.scss
@@ -87,22 +87,8 @@ $confval-dt-border: #666666;
 //
 // Code Colors
 // --------------------------------------------------
-$code-color:        hsl(350, 100%, 40%);
-//
-$code-color-bash:   hsl(0, 100%, 0%);
-$code-color-css:    hsl(275, 100%, 40%);
-$code-color-file:   hsl(328, 100%, 50%);
-$code-color-html:   hsl(275, 100%, 40%);
-$code-color-js:     hsl(128, 100%, 20%);
-$code-color-method: hsl(210, 100%, 35%);
-$code-color-php:    hsl(210, 100%, 30%);
-$code-color-py:     hsl(210, 100%, 50%);
-$code-color-rst:    hsl(175, 100%, 15%);
-$code-color-shell:  hsl(0, 100%, 0%);
-$code-color-sql:    hsl(175, 100%, 15%);
-$code-color-typoscript: hsl(12, 100%, 20%);
-$code-color-xml:    hsl(275, 100%, 40%);
-$code-color-yaml:   hsl(170, 80%, 25%);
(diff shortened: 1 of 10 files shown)


brotkrueml:
Hmm, personally I liked the colors. The icons make the reading harder (at least for me). But maybe I am too old for such things.

What does the icons `A2` and `M(arrow down)` mean?

linawolf:
M arrow down is the official logo for markdown. The sign with the 5 is the official logo of HTML5, The wrench is for configuration with json, which has no offical logo. 

The icons would repeat in both code-blocks and inline-code so you can always make the connection and get used to the symbols. The colors had no meaning really and it was hard to impossible to figure out what language they refer to?

Whereever it does not really matter we can use a simple literal which does not have a logo:

image

```
`literal`
```

But I think when we explicitly specify the language of inline code we should also display it...

brotkrueml:
I am not convinces that this is the way, especially I on't like this icon:

image

Yes, I know, it is the official PHP one. But it is hard to see (in comparison with the other icons).

For XML, why that icon?

image

For me it is a "Git fork" icons.

Hmm, I would like to see another variant, like just using plain CSS "PHP", "JS", "XML", "JSON", "TYPOSCRIPT", etc - like with some background colour and white text colour. This can be done with CSS "::before" and "content".

garvinhicking:
I found the old colors to be non-clear by themself too, and can think of icons being an improvement.

However, it does look quite crowded then in inline markup. And yes, the xml, php and json icons do look a bit off. Maybe having xml and php spelled out would be more clear. 

Json does have a logo https://en.m.wikipedia.org/wiki/JSON but that doesnt really look unique to me, and I don't associate it with json.

How about maybe making this toggle-able with a cookie/localstorage js functionality? At some point we might get dark/light-mode toggles too so that could be grouped.

I'm quite torn, looking at it I prefer one or the other just by chance 🤔

brotkrueml:
FYI: dark/light mode will be no toggle, but a media query which takes the current browser/os setting into account.

Toggles won't work, rather no one will use them. We have already toggles in the Sphinx-based theme, have you encountered them?

I think, either we provide a way, which is good, or we don't implement it. Letting the user decide is bad UI/UX in this case IMHO. Looks like, yeah, we implemented something, but we think it is not really good, so you can deactivate it. 

I will provide a separate PR with another approach.

garvinhicking:
IMO it is good style to offer a toggle and just inherit the default.

Some prefer dark mode in OS but reading docs in light mode.

I think this should not be pure CSS.

I like choices. One thing I hate about Apples "one size fits all" approach.

sarahmccarthy123:
I quite like the new style with the icons. Obviously it is crowded because you are showing all the examples so usually it won't look like that. The old style looks too faded for me. I think icons and infographics generally are the future, but that's just me... :-)

PR #312, opened 2024-01-21
[TASK] Add asset handling to makefile

linawolf:
For now it only runs on ddev.


brotkrueml (approved):

brotkrueml:
Can also run locally, if node and npm is installed.

linawolf:
Well I cannot do that on my current machine. If you know how to add this to the makefile go ahead ;-)

PR #319, opened 2024-01-26
[TASK] Add Basic Styling for Code-Block Caption

linawolf:
We can discuss in a follow up weather we want to include additional information like the icons proposed here: https://github.com/TYPO3-Documentation/render-guides/pull/310

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/components/_code.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_code.scss
@@ -24,8 +24,14 @@
     padding: .75rem;
     margin-bottom: 1rem;
     .code-block-caption {
-        padding-bottom: .75rem;
-        margin-bottom: .75rem;
+        .caption-text {
+            display: inline-block;;
+            background: $white;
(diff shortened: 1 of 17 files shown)


brotkrueml (commented):
For me it is too less contrast and distinction between the caption and the code:

Bildschirmfoto vom 2024-01-26 17-55-23

Would like to have a better separation.

linawolf:
We could use the url-bar like background from my other PR again:

image

brotkrueml:
Hmm, contrast still too less.

Maybe we should use a dark separator line below the caption. But I assume that won't look good.

garvinhicking:
I agree, this is a bit too light and doesn't set off enough. How about either a clear dark border, or a dark box with light text (inverted so to say)? I do think box caption could structurize the perception of sections so clear lines or boxes would be good?

linawolf:
With a shadow maybe?

image

garvinhicking:
(On it with an idea on how to proceed)

linawolf:
solved in a follow up

Issue #316, opened 2024-01-25
project-contact not considered


brotkrueml:
The project contact is always linked to the Docs Team page on typo3.org, regardless of the configuration in project-content attribute of the extension tag in guides.xml.

- [ ] Consider project-contact attribute when it is defined
- [ ] Do not display the Contact link in the footer when it is not defined or empty

PR #325, opened 2024-01-27
[FEATURE] Display Project Info from guides.xml in Footer

linawolf:
Resolves https://github.com/TYPO3-Documentation/render-guides/issues/316
Depends on https://github.com/TYPO3-Documentation/render-guides/pull/326

### DIFF
--- a/Documentation/guides.xml
+++ b/Documentation/guides.xml
@@ -13,6 +13,10 @@
         edit-on-github="TYPO3-Documentation/render-guides"
         edit-on-github-branch="main"
         interlink-shortcode="t3docsRenderGuides"
+        project-contact="https://typo3.slack.com/archives/C0638JMJVLY"
+        project-home="https://github.com/TYPO3-Documentation/render-guides"
+        project-issues="https://github.com/TYPO3-Documentation/render-guides/issues"
+        project-repository="https://github.com/TYPO3-Documentation/render-guides"
         typo3-core-preferred="stable"
     />
     <!-- explicitly link to stable versions of extensions -->
--- a/packages/typo3-docs-theme/resources/template/structure/layoutParts/footer.html.twig
+++ b/packages/typo3-docs-theme/resources/template/structure/layoutParts/footer.html.twig
@@ -3,11 +3,22 @@
     <div class="frame frame-ruler-before frame-background-dark">
         <div class="frame-container">
             <div class="frame-inner">
-                <ul class="footer-simplemenu">
-                    <li><a href="https://typo3.org/community/teams/documentation/#c9886" title="Contact"><span>Contact</span></a></li>
-                    <li><a href="https://github.com/TYPO3-Documentation/DocsTypo3Org-Homepage/issues" rel="nofollow noopener" title="Issues"><span>Issues</span></a></li>
-                    <li><a href="https://github.com/TYPO3-Documentation/DocsTypo3Org-Homepage" rel="nofollow noopener" title="Repository"><span>Repository</span></a></li>
-                </ul>
+                {% if env.projectNode.variables.project_home or env.projectNode.variables.project_contact or env.projectNode.variables.project_issues or env.projectNode.variables.project_repository -%}
+                    <ul class="footer-simplemenu">
+                        {% if env.projectNode.variables.project_home -%}
+                            <li><a href="{{ renderNode(env.projectNode.variables.project_home) }}" title="Home" rel="nofollow noopener"><span>Home</span></a></li>
+                        {% endif %}
+                        {% if env.projectNode.variables.project_contact -%}
+                            <li><a href="{{ renderNode(env.projectNode.variables.project_contact) }}" title="Contact" rel="nofollow noopener"><span>Contact</span></a></li>
+                        {% endif %}
+                        {% if env.projectNode.variables.project_issues -%}
+                            <li><a href="{{ renderNode(env.projectNode.variables.project_issues) }}" rel="nofollow noopener" title="Issues"><span>Issues</span></a></li>
+                        {% endif %}
+                        {% if env.projectNode.variables.project_repository -%}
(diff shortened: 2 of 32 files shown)


Documentation/guides.xml
```diff
@@ -14,6 +14,10 @@
         edit-on-github-branch="main"
         interlink-shortcode="t3docsRenderGuides"
         typo3-core-preferred="stable"
+        project-home="https://github.com/TYPO3-Documentation/render-guides"
```
brotkrueml, line 17: Until now, the attributes are sorted alphabetically. Can we stay that way?

packages/typo3-docs-theme/resources/template/structure/layoutParts/footer.html.twig
```diff
@@ -3,11 +3,22 @@
     <div class="frame frame-ruler-before frame-background-dark">
         <div class="frame-container">
             <div class="frame-inner">
-                <ul class="footer-simplemenu">
-                    <li><a href="https://typo3.org/community/teams/documentation/#c9886" title="Contact"><span>Contact</span></a></li>
-                    <li><a href="https://github.com/TYPO3-Documentation/DocsTypo3Org-Homepage/issues" rel="nofollow noopener" title="Issues"><span>Issues</span></a></li>
-                    <li><a href="https://github.com/TYPO3-Documentation/DocsTypo3Org-Homepage" rel="nofollow noopener" title="Repository"><span>Repository</span></a></li>
-                </ul>
+                {% if env.projectNode.variables.project_home or env.projectNode.variables.project_contact or env.projectNode.variables.project_contact or env.projectNode.variables.project_repository -%}
```
brotkrueml, line 6:
```suggestion
                {% if env.projectNode.variables.project_home or env.projectNode.variables.project_contact or env.projectNode.variables.project_issues or env.projectNode.variables.project_repository -%}
```

packages/typo3-docs-theme/src/EventListeners/AddThemeSettingsToProjectNode.php
```diff
@@ -0,0 +1,26 @@
...
+{
+    public function __construct(
+        private readonly Typo3DocsThemeSettings $themeSettings,
+    ) {}
+
+    public function __invoke(PostProjectNodeCreated $event): void
+    {
+        $projectNode = $event->getProjectNode();
+        foreach ($this->themeSettings->getAllSettings() as $key => $setting) {
+            if ($setting !== '') {
```
brotkrueml, line 21: Maybe a trim around `$setting`. It is an edge case.

packages/typo3-docs-theme/src/EventListeners/AddThemeSettingsToProjectNode.php
```diff
@@ -0,0 +1,26 @@
...
+                $projectNode->addVariable($key, new PlainTextInlineNode(strval($setting)));
```
brotkrueml, line 22: You check for an non-empty string above, but then cast it to a string with strval(). Either the check should be adjusted (maybe null?) or the strval is superfluous.

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
```diff
@@ -23,4 +23,12 @@ public function getSettings(string $key, string $default = ''): string
         }
         return $this->settings[$key];
     }
+
+    /**
+     * @return array<string, string>
+     */
+    public function getAllSettings(): array
```
brotkrueml, line 30: Can we have unit tests for this class (as you introduce a new method)?

brotkrueml (commented):
Is the patch about "contact info" or "project infos"?

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
linawolf: You want unit tests for the getters and setters on a data object? I find that quite superfluous

linawolf:
> Is the patch about "contact info" or "project infos"?

I dont really see the difference, which one do you suggest?

packages/typo3-docs-theme/src/EventListeners/AddThemeSettingsToProjectNode.php
linawolf: true, even though the $settings where supposed to be `array<string, string>` still a boolean value was passed. (wondering why phpstan didnt catch this). Anyway, I fix it now

brotkrueml:
> > Is the patch about "contact info" or "project infos"?
> 
> I dont really see the difference, which one do you suggest?

Well, the project tag has 

> > Is the patch about "contact info" or "project infos"?
> 
> I dont really see the difference, which one do you suggest?

Well, we have contact, home, issues, repository, so I would prefer project as the umbrella.

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
brotkrueml: You have logic in it (isset check). But I am fine to move that to a dedicated PR.

Documentation/guides.xml
linawolf: Ack

packages/typo3-docs-theme/resources/template/structure/layoutParts/footer.html.twig
linawolf: Good catch!

packages/typo3-docs-theme/src/Settings/Typo3DocsThemeSettings.php
linawolf: Feel free to provide a PR

linawolf:
@brotkrueml can we merge this?

brotkrueml (approved):

Issue #288, opened 2024-01-15
Wrong display of the list of release versions


FlorianT03:
When I click on the "+" of "Introduction" on the left, for example, a list of all the release versions for the docs opens automatically. However, this should not happen, only the sub-items for the introductions should appear; this also applies to the other items such as "Quick start" etc.

https://docs.typo3.org/p/georgringer/news/draft/en-us/Index.html

PS: Apparently a single click is enough to display the list of releases, I accidentally clicked on an empty part of the website and the list of releases appeared.

brotkrueml:
Yup, this is known. The click handler seems to be registered to the whole document instead of only the version selector.

PR #339, opened 2024-02-11
[BUGFIX] Use correct selector for version

brotkrueml:
Resolves: #288


linawolf (approved):
Looks good and works locally

Issue #344, opened 2024-02-16
Searchbox on top of document is missing


linawolf:

brotkrueml:
We agreed some time ago that it is not necessary anymore, as Elasticsearch will cover the search per document in the future (at least that was my understanding).

linawolf:
The searchbox on the top of the document is leading to elastic search ;-)

PR #349, opened 2024-02-17
[TASK] Use search element

brotkrueml:
It is intended to also provide the `role="search"` attribute: Not all assistive technologies may recognize the new HTML element by now.

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/search
See: https://caniuse.com/mdn-html_elements_search
Related: #344


linawolf:
Cool, didnt know about this one!

linawolf (approved):

Issue #348, opened 2024-02-17
Handle Localizations


linawolf:
In some manuals like getting started we feature translations. These are in a subfolder called "Localization.xx_YY". They need to be treated as independent manuals and not be rendered within the main manual

PR #351, opened 2024-02-17
[BUGFIX] Ignore localization folders

linawolf:
resolves https://github.com/TYPO3-Documentation/render-guides/issues/348

### DIFF
--- a/packages/typo3-docs-theme/src/EventListeners/IgnoreLocalizationsFolder.php
+++ b/packages/typo3-docs-theme/src/EventListeners/IgnoreLocalizationsFolder.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace T3Docs\Typo3DocsTheme\EventListeners;
+
+use phpDocumentor\Guides\Event\PostCollectFilesForParsingEvent;
+use phpDocumentor\Guides\Files;
+
+final class IgnoreLocalizationsFolder
+{
+    /**
+     * @see https://regex101.com/r/zUNAFQ/1
+     */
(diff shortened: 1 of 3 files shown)


packages/typo3-docs-theme/src/EventListeners/IgnoreLocalizationsFolder.php
```diff
@@ -0,0 +1,25 @@
...
+
+use phpDocumentor\Guides\Event\PostCollectFilesForParsingEvent;
+use phpDocumentor\Guides\Files;
+
+final class IgnoreLocalizationsFolder
+{
+    /**
+     * @see https://regex101.com/r/zUNAFQ/1
+     */
+    private const LOCALIZATION_FOLDER_REGEX = '/^Localization\\.[a-z]+_[A-Z]+/s';
```
brotkrueml, line 13: The regex101 and this regex is not in sync:
regex101 is `^Localization\.[a-z]{2}\_[A-Z]{2}`

This only covers the usual language/country combinations. But there are others, like `zh_Hans_CN`

See also:
- https://www.ietf.org/rfc/rfc4646.html#appendix-B
- https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Localization/Locales.php#L103

I suggest to add it to the regex reference, either we support it or not - but it should be documented that way.

linawolf: I also left a todo comment, we have to consider several places if we want to allow more formats.

brotkrueml (approved):

PR #356, opened 2024-02-20
[FEATURE] Add list of available languages for code blocks

brotkrueml:


brotkrueml:
This is a draft for outputting the available languages for a code-block directive.

This is meant for usage in
https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Code/Codeblocks.html#available-lexers

It should be used only internal in our docs.

@linawolf What do you think?

linawolf:
@brotkrueml I like the idea!

brotkrueml:
@linawolf Then now it can be reviewed :-)

tests/Integration/tests/codehighlight-languages/input/index.rst
```diff
@@ -0,0 +1,7 @@
+===============================
+Registered code-block languages
+===============================
+
+All registered languages for code blocks:
+
+..  codeblock-languages::
```
linawolf, line 7: I would suggest that we use a common domain for directives that are used to document our rst-implementation. Maybe 

rest:codeblock-languages

Ist seems like we are gonna have more then just one

brotkrueml: This is valid?

    .. rest:codeblock-languages::

How do I name then the according template respective the path?

linawolf: just like we do in the php domain?

linawolf: directive/rest/codeblock-languages.html.twig

brotkrueml: Why "rest"? Maybe "t3" would be better as it is more an internal directive?

brotkrueml: Or "internal"? That's make the purpose clear.

linawolf: h2document?

garvinhicking: OMG, is "h2document" = "how to document"? This is so not clear :-P

I would vote for "guides", "t3", "internal" in that order...

linawolf: I like guides. I would reserve t3 for typo3 specific stuff, like if we ever want to do something expecially for TypoScript or whatever?

brotkrueml: Okay, will use guides :1st_place_medal:

brotkrueml: Done, please have a look.

garvinhicking (approved):
Very cool!

Next up, `guides:inventories`? 😇😶‍🌫️

linawolf (approved):

PR #360, opened 2024-02-22
[TASK] Adjust font size of inline code

brotkrueml:
The font size of inline code (for example, `:file:`, `:php:`, etc) is too small in comparison with the surrounding text. It is now increased for better readability.

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/components/_code.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_code.scss
@@ -90,7 +90,7 @@
  */
 .code-inline {
     font-family: $font-family-monospace;
-    font-size: 75%;
+    font-size: 90%;
     border-radius: $border-radius;
     border: 1px solid #fff;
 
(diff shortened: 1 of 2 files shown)


brotkrueml:
Before:

Bildschirmfoto vom 2024-02-22 20-17-25

After:

Bildschirmfoto vom 2024-02-22 20-17-50

brotkrueml:
> This always bothered me, it looks as if the bottom baseline of the text is not aligned with the inline styles, but I never properly investigated and applied a ruler on it. ;)
> 
> I'd love if this could be improved, the new screenshot looks improved in that regard, but still as if there's a pixel difference?

The baseline is the same (tested with a ruler), for the "normal" text and the inline code (before and after). Visually (at least for me), it is on the same baseline.

brotkrueml:
> I think it currently matches the size of the text and afterwards will look bigger then the surrounding text.

Sorry, I don't get the point?

linawolf:
I think my feeling that the monospaced text is now bigger then the normal text stems from the letters beeing wider. That was probably the reason why someone made it initially smaller. But as the code is usually very important to us I am ok with this change

linawolf (approved):

garvinhicking:
Yes! I checked the baseline. This is great. I do see Linas point about the a different perception, because of the difference in the font. But to me this looks much more "on a line" now, thank you!

garvinhicking (approved):

PR #374, opened 2024-03-02
[FEATURE] List available standard inventories

linawolf:
They can be documented by using directive ..  guides:inventories::

Resolves https://github.com/TYPO3-Documentation/render-guides/issues/373

### DIFF
--- a/packages/typo3-docs-theme/resources/template/body/directive/guides/inventories.html.twig
+++ b/packages/typo3-docs-theme/resources/template/body/directive/guides/inventories.html.twig
@@ -1,0 +1,5 @@
+<ul>
+    {% for inventory in getStandardInventories() %}
+        <li>{{ inventory }}</li>
+    {% endfor %}
+</ul>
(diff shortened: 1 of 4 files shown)


brotkrueml (commented):
Is there somehow the possibility to describe the inventories? For example, "t3install" is the "Upgrade Guide" and not the "Installation Guide". Maybe a method in the enum which returns the descriptions?

linawolf:
It is possible if we have the descriptions in the enum. Or we could just fetch all the inventories and get the title of the "start" reference...

brotkrueml:
As the inventories are already explicitely defined in the enum I think defining a method would be the easiest way. Otherwise all repos must by fetched by suspicion upon rendering, or only when the directive is used - both has an unnecessary overhead as this directive is only used in ony place, I assume.

garvinhicking (approved):
Very cool, thanks! We could address descriptions in a follow-up? That's a good idea too!

PR #383, opened 2024-03-04
[TASK] Style changes

linawolf:
These style changes were suggested by @rfoucard
who asked me to commit them for her.


packages/typo3-docs-theme/assets/sass/_text.scss
```diff
@@ -1,7 +1,25 @@
-
+a {
+    color:inherit;
+    text-underline-offset: 25%;
+    &:hover {
+        text-decoration: none;
```
brotkrueml, line 5: I am not convinced that a hover should remove the underline - I think there may be better ways for signalling hover. Additionally, I miss a :focus for keyboard users.

packages/typo3-docs-theme/assets/sass/_text.scss
```diff
@@ -1,7 +1,25 @@
...
+    }
+}
 article {
     a:not([class*="btn"]) {
         text-decoration: underline;
+	    color: $secondary;
+        &:hover {
+            text-decoration: none;
```
brotkrueml, line 13: Here too.

brotkrueml:
Reviewing ...

packages/typo3-docs-theme/assets/sass/_text.scss
```diff
@@ -1,7 +1,25 @@
...
+        }
+    }
+    p, li {
```
brotkrueml, line 16: The font size is not adjusted at all places:

before:
Bildschirmfoto vom 2024-03-04 18-36-12

after:
Bildschirmfoto vom 2024-03-04 18-36-58

packages/typo3-docs-theme/assets/sass/_text.scss
brotkrueml: [links-2024-03-04_18.40.09.webm](https://github.com/TYPO3-Documentation/render-guides/assets/2566282/54ca4a8c-c0b3-43fc-b058-bb871f297d6c)

packages/typo3-docs-theme/assets/sass/_text.scss
brotkrueml: I think, if we really change the font size we should do this at a more central place.

Additionally, for me the different font sizes looks strange:
- The font size of the content is larger
- Navigation stays smaller

packages/typo3-docs-theme/assets/sass/components/_code.scss
```diff
@@ -90,63 +93,8 @@
  */
 .code-inline {
     font-family: $font-family-monospace;
-    font-size: 90%;
     border-radius: $border-radius;
     border: 1px solid #fff;
-
-    &.bash {
```
brotkrueml, line 97: Then we should switch to use `:code:` generally and also recommend it on migration.

packages/typo3-docs-theme/assets/sass/components/_pagination.scss
```diff
@@ -0,0 +1,10 @@
+.page-link {
+    color: inherit;
+    background: #f2f2f2;
+    &:hover {
```
brotkrueml, line 4: We are using scss, so can be written as:

    &:hover,
    &:focus  {
         color: inherit;
    }

Then the property is only in one place.

packages/typo3-docs-theme/assets/sass/components/_tabs.scss
```diff
@@ -1,6 +1,21 @@
+.tabs {
```
brotkrueml, line 1: I am not convinced that the tabs content should be that prominent with the gray background:

Bildschirmfoto vom 2024-03-04 18-51-12

brotkrueml: Additionally, a margin bottom is missing.

packages/typo3-docs-theme/assets/sass/_text.scss
brotkrueml: Maybe we should also largen the content width?

packages/typo3-docs-theme/assets/sass/components/_code.scss
brotkrueml: Additionally, the node.class should not be necessary anymore?
https://github.com/TYPO3-Documentation/render-guides/blob/main/packages/typo3-docs-theme/resources/template/inline/textroles/code.html.twig

packages/typo3-docs-theme/assets/sass/_text.scss
linawolf: How about we change the color of the underline to TYPO3 orange on hover?

packages/typo3-docs-theme/assets/sass/_text.scss
linawolf: Should we stick with one font size maybe?

packages/typo3-docs-theme/assets/sass/_text.scss
garvinhicking: If that is okay for https://github.com/rfoucard I think that would be good! Let's change the highlight color to orange, and keep the underline, and go from there. We should go an iterative way with constant improvements, else won't get far at all.

To make people happily check out the new rendering, we need people not to be disturbed by little things, which we notice now. Let's do this. :)

packages/typo3-docs-theme/assets/sass/components/_tabs.scss
garvinhicking: IMO this is better than the currently too much detached tabs list, but I agree on smaller tabs it looks hefty. Maybe if we can add the margin bottom and try out further ideas after that?

packages/typo3-docs-theme/assets/sass/components/_code.scss
garvinhicking: IMO the whole "icon design" is up for discussion at a later place, so that we may use hover or css :after(...) features to add "PHP", "YAML" etc. to some kind of box. IMO we can remove the coloring at this point because it needs its own treatment, and currently the colors are really detracting from reading....?

packages/typo3-docs-theme/assets/sass/_text.scss
rfoucard: Hi,
another way can be to change the text-decoration-thickness to 2px, it avoid using only color to give information and it is better than removing the underline.

packages/typo3-docs-theme/assets/sass/_text.scss
rfoucard: Indeed the font size should be the same in the central column, but it doesn't need to be the same in the navigation part like the aside menu or the breadcrumb. There is no accessibility rule saying that the font-size must be unique in an interface.

packages/typo3-docs-theme/assets/sass/components/_pagination.scss
linawolf: regarding accessibility focus and hover should not be the same, but focus should have a general styling like a box or some such?

garvinhicking:
@brotkrueml and I will take some time later today to separate the "discussable" parts of this PR into a seperate one, and merge that what we can all agree on :-)

packages/typo3-docs-theme/assets/sass/components/_tabs.scss
linawolf: The tabs also don't really convince me, but we can also solve them in a follow up

packages/typo3-docs-theme/assets/sass/_text.scss
linawolf: Good idea, I will give it a try!

packages/typo3-docs-theme/assets/sass/components/_pagination.scss
rfoucard: @linawolf you're right. It would be interesting to create a separated pull request "focused on focus". We could give the outlined style to all focused link in a more global way.

packages/typo3-docs-theme/assets/sass/_text.scss
linawolf: Done

packages/typo3-docs-theme/assets/sass/_text.scss
linawolf: done

packages/typo3-docs-theme/assets/sass/components/_tabs.scss
linawolf: lets make a  follow up for the tab styles?

packages/typo3-docs-theme/assets/sass/components/_pagination.scss
linawolf: In my adjustment I was now also tackling the focus. We have to have a look at the main menu focus-wise but should do that in a follow up

packages/typo3-docs-theme/assets/sass/_text.scss
brotkrueml: I am not speaking about accessibility. I am speaking that this looks not good to me, "unround" would I say in German.

packages/typo3-docs-theme/assets/sass/components/_code.scss
brotkrueml: Then we should tackle this beforehand. Otherwise we encourage people to use `code` as there is no difference visible.

packages/typo3-docs-theme/assets/sass/_text.scss
```diff
@@ -1,7 +1,25 @@
-
+a {
+    color:inherit;
+    text-underline-offset: 25%;
+    &:hover, &:focus {
+        text-decoration-thickness: 2px;
+    }
+}
+*:focus-visible {
+    @include focus();
```
brotkrueml, line 9: What problem is solved with this?

packages/typo3-docs-theme/assets/sass/components/_button.scss
```diff
@@ -2,6 +2,9 @@
     display: inline-flex;
     align-items: center;
     text-align: left;
+    &:focus-visible {
```
brotkrueml, line 5: Why is this defined here, isn't focus-visible from above already applied here?

packages/typo3-docs-theme/assets/sass/components/_tabs.scss
brotkrueml: I think, just marking the surrounding border darker should be enough.

brotkrueml (commented):
Some remarks from me again.

I suggest to split this PR to separate ones by scope:
- Links
- Tabs
- "page-link"?
- Font size
- "focus-visible"?
- Inline code
- ...

Currently, I don't get which styles are applied for which purpose. So, I can't really review and test it accordingly.

brotkrueml:
Adminitions are already extracted in #385

packages/typo3-docs-theme/assets/sass/_text.scss
linawolf: If you currently use tab on the documentation page you cannot see where the focus is. Try using Tab on the new rendering, the focus is clear and the keyboard can be used for navigation. This is a big step toward A11y

linawolf:
Font styles and weights extracted to #387

linawolf:
Fix top search box extracted to #389

linawolf:
Extract keyboard focus to #388

linawolf:
solved in a follow up

PR #385, opened 2024-03-05
[TASK] Restyle admonitions

brotkrueml:
Additionally, danger and error have been missing in the template.

Related: #383


garvinhicking:
We should try to progressively address each style approvement, this would be the most undisputed change for now :)

garvinhicking (approved):

linawolf (approved):

PR #387, opened 2024-03-06
[TASK] Improve Font weight, styles and colors

linawolf:
Before:

image

After:

image


brotkrueml (commented):
As I already said, I am not convinced about this change.

Currently, less content is displayed, so the content frame should be wider.

Additionally, IMHO also the navigation should have the same font size (with a wider content block this should then be okay).

Also, on mobile (small screens) the font-size should be smaller then again. Has to be tested.

linawolf:
This is solved by smaller tickets by rachel

PR #388, opened 2024-03-06
[FEATURE] Improve visibility of keyboard focus

linawolf:
This enhances the accessibility of the docs. Currently on most elements the focus is not visible at all (menu) or poorly visible (buttons).

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/components/_button.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_button.scss
@@ -2,6 +2,12 @@
     display: inline-flex;
     align-items: center;
     text-align: left;
+    &:focus-visible {
+        // Override the default btn styles from bootstrap as they are poorly visible
+        @include focus();
+        outline: auto;
+        outline-color: $secondary-dark;
+    }
 }
 .btn-icon {
     &:first-child {
--- a/packages/typo3-docs-theme/assets/sass/layout/navigation/_main_navigation.scss
+++ b/packages/typo3-docs-theme/assets/sass/layout/navigation/_main_navigation.scss
@@ -37,6 +37,10 @@
         padding: calc(#{$spacer} / 4) 0;
         padding-right: 1.5em;
         text-decoration: none;
+        &:focus-visible {
+            color: $white;
+            background:$secondary-dark;
+        }
     }
     ul {
         padding-left: $spacer;
@@ -85,6 +89,11 @@
             &:before {
                 display: none;
             }
+        }
+
+        &:focus-visible {
+            color: $white;
+            background:$tertiary;
         }
     }
     .current > a {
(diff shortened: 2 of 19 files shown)


packages/typo3-docs-theme/assets/sass/components/_button.scss
```diff
@@ -2,6 +2,12 @@
     display: inline-flex;
     align-items: center;
     text-align: left;
+    &:focus-visible {
+        // Override the default btn styles from bootstrap as they are poorly visible
+        @include focus();
+        outline: auto;
```
brotkrueml, line 8: This should not be necessary, as `outline: auto` is a shortcut for `outline-color: auto` which is overwritten below.
https://developer.mozilla.org/en-US/docs/Web/CSS/outline#formal_definition

packages/typo3-docs-theme/assets/sass/layout/navigation/_main_navigation.scss
```diff
@@ -37,6 +37,10 @@
         padding: calc(#{$spacer} / 4) 0;
         padding-right: 1.5em;
         text-decoration: none;
+        &:focus-visible {
+            color: $white;
```
brotkrueml, line 41: Why using a background-color and not just the common outline?

Additionally, the current solution looks not really good:
mainnav

- text should have a padding left and right
- The + symbol is not really visible anymore

packages/typo3-docs-theme/assets/sass/mixins/_focus.scss
```diff
@@ -0,0 +1,4 @@
+@mixin focus() {
+    box-shadow: $focus-box-shadow;
```
brotkrueml, line 2: There seems to be a glitch, I see a black border top (Firefox):

focus

Also Chromium:
focus2

packages/typo3-docs-theme/assets/sass/layout/navigation/_main_navigation.scss
linawolf: as I already wrote in the other PR, because I was not able to do it better. The plus does not work with the keyboard anyway. The menu needs a makeover which someone :tm: has to do.

packages/typo3-docs-theme/assets/sass/components/_button.scss
linawolf: If you don't do it the box-shadow is not visible.

packages/typo3-docs-theme/assets/sass/mixins/_focus.scss
linawolf: whatever, make a follow up

linawolf:
Solved in a folllow up

PR #389, opened 2024-03-06
[BUGFIX] Fix display of top search

linawolf:
Before:

2024-03-06 06_30_04-Developers — Render guides documentation

After:

image


brotkrueml (commented):
Why are the two font files added?

linawolf:
because they crept in. deleted them

brotkrueml (approved):

Issue #390, opened 2024-03-07
Edit on GitHub Button does not work in AdminPanel


linawolf:
https://docs.typo3.org/c/typo3/cms-adminpanel/main/en-us/Index.html

Links to 
https://github.com/typo3/typo3/edit/main/Documentation/Index.rst

garvinhicking:
This is surely because the edit-on-github links were meant to work on the root level of an extension or github repository. The link parts:

```
https://github.com/ (fixed)
typo3/typo3 (from <project.edit-on-github>)
/edit (fixed)
/main (from <project.edit-on-github-branch>)
/Documentation/Index.rst (directory relative from the render root)
```

I'm afraid I don't think we have the ability to get the missing directory parts (`typo3/sysext/adminpanel/`) from a variable of guides.xml, it's simply not known to the rendering.

So for core manuals we need a new variable like `<project.edit-on-github-basedir>` which would by default be set to an empty string, and in the case of `cms-adminpane` set to `/typo3/sysext/adminpanel`.

brotkrueml:
Don't we have some PHP code where we can check for the repository name and if it starts with typo3/cms-* we adjust the link?

garvinhicking:
The rendering is done only with the context of "Documentation/", I think this is what the github action checks out and passes to the rendering container. So the rendering container does not know of its directory structure outside of the Root where "Documentation/" is in.

I think we can't even check the composer.json in the root path, because only "Documentation/" is contained in the directory structure within the container....

linawolf:
There was a setting in the old settings.php to influence the path on the edit on github button. That setting was not migrated or used by us. Making a PR

linawolf:
Resolved

PR #393, opened 2024-03-07
[BUGFIX] Edit on GitHub Button with adjusted path

linawolf:
resovles #390


garvinhicking (approved):
Wow! That was insanely fast, thank you. Looking good to me, also thanks for taking the migration into account already!

PR #397, opened 2024-03-08
[TASK] Add "t3upgrade" key to the default inventories

brotkrueml:
The `t3install` references the "Upgrade Guide" which is semantically wrong. Therefore, the `t3upgrade` key has been introduced.

As both are available in the wild, both are valid, but `t3install` should be considered legacy.


garvinhicking (approved):
I think it is good to try to clean this up. We probably cannot change "https://docs.typo3.org/m/typo3/guide-installation/main/en-us/" to "guide-upgrade" without breaking stuff, I guess? I always stumble across this difference and what is what.

linawolf:
I would also agree on moving the guide and making a redirect. However then we can easaly change the links here, so we can do that in another PR

linawolf (approved):

PR #399, opened 2024-03-09
[BUGFIX] Fix interlink version fallback

### LINKED ISSUE #396: No default interlink version fallback when `Typo3DocsThemeExtension.typo3-core-preferred` not set
guides.xml allows to set this:

```
<extension class="\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension" 
  typo3-core-preferred="stable"/>
```

to indicate the version that links shall go to.

However, when that attribute is not set, the rendering fails with wrong interlinks, as you can see for example here:

https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/actions/runs/8189596042/job/22394688950?pr=402#step:3:284

```
[2024-03-07T14:18:20.589187+00:00] app.WARNING: Inventory group with key "std:label" (std-label) 
not found in inventory [https://docs.typo3.org/m/typo3/tutorial-getting-started//en-us/.]
(https://docs.typo3.org/m/typo3/tutorial-getting-started//en-us/)  {"rst-
file":"Maintainers/NewMajorCoreVersion","type":"ref","targetReference":"installation-ddev-
tutorial","value":"","linkType":"std:label","interlinkDomain":"t3start"} []
```

Note that the link is formatted like `/m/typo3/tutorial-getting-started//en-us/` but it should be `/m/typo3/tutorial-getting-started/main/en-us/` instead (by default?)

This makes tests fail in all repositories that check for render warnings and use interlinks without a specifically defined inventory.

linawolf:
Add a test without typo3-core-preferred in guides.xml

Move interlink tests in common directories, so they are easier found.

resolves #396


garvinhicking (approved):

garvinhicking:
Nice, this looks good!! Thanks for getting on it so quickly :)

PR #412, opened 2024-03-12
[TASK] Change ordered list marker color to text color

fe-breuer:


garvinhicking:
Our main branch was out of sync so we could not easily commit the generated CSS files too. We can commit the generated CSS files after that, when the other PR https://github.com/TYPO3-Documentation/render-guides/pull/411 is merged.

PR #440, opened 2024-03-16
[TASK] Remove raw directive from rendertest

linawolf:
we do not support it for security reasons


garvinhicking:
Hm, isn't the note then valid, and it would be good to have it in just to see that it's not rendered? But it could be shortened...?

brotkrueml (commented):
I assume, as there is no "raw" directive in guides available, this page can be removed safely.

brotkrueml (approved):

linawolf:
@garvinhicking I think this should go in an automated test, made a ticket: https://github.com/phpDocumentor/guides/issues/942

PR #462, opened 2024-03-22
[TASK] Allow searching in single manual scope

soee:
This commit extends the documentation search form with a new select field where user can choose if he wants to search either in the whoel documentation or only single one - the one he sees atm.

The single manual scope is determined by its path (or slug). For example, if the user is on this page:
/m/typo3/tutorial-getting-started/main/en-us/Installation/Install.html the scope will be `m/typo3/tutorial-getting-started/main/en-us` and this value is passed through the search form to the search results page.

This scope value is added to the search form by the JavaScript code, which checks the `window.location.pathname` value and matches it against regular expression pattern to verify if it points to the one of the supported documentation paths.

There were also small updates to styles and search form structure on small screens.


linawolf (approved):
Thanks for these changes

PR #478, opened 2024-03-31
[TASK] styles text contrast

rfoucard:
Hello,

This pull request changes globally in the css the variable $gray-800 from #666666 to #595959. 
It will be subtile and make all the dark grays (colors and backgrounds) a little bit darker. 
The real benefit is that it will give the overall documentation text a better contrast, (7.1) valid with WCAG AAA level.


garvinhicking (approved):
Can't easily see the results of this without a screenshot or sample page, but I fully trust your recommendation😇 Thank you!

linawolf:
For the record, After:
image

Before:
image

linawolf:
I cannot really see a difference, but I guess it doesn't hurt?

PR #483, opened 2024-04-02
[TASK] Add SQL popover highlighting

garvinhicking:
We missed `:sql:` role handling to be specific, and receive a distinct popover.


linawolf (commented):
Thanks for catching this, shouldnt we add an alias for "mysql?"

garvinhicking:
> Thanks for catching this, shouldnt we add an alias for "mysql?"

I shortly thought about this, but then we'd also need to alias sqlite and pgsql and postgresql and postgres and mariadb too probably. So let's just only implement "sql" to not have confusion about it?

linawolf:
I think mysql was possible in the old rendering container and was used for historic reasons. so I would just add it as an alias that doesnt hurt but not advertise it

PR #484, opened 2024-04-02
[BUGFIX] Remove duplicate definitions of Fluid and PHP text roles

garvinhicking:
Both definitions look duplicated to me, can you check @linawolf ?


linawolf (approved):
I think it was a merging artifact

PR #498, opened 2024-04-10
[DOCS] Add list of inventories with descriptions

garvinhicking:
Previously I could look into the settings.cfg to see keys and names of interlink inventories (because I do not know them all by heart), but now this is only available in the render-guides repository.

There seems to be a twig directive to get the enum values and their links, but I acutally need more:

* The name of the inventory as rendered for the Title
* The actual URL used
* The key of the inventory

So I made a manual list of this because I didn't want to enhance the ENUM to do too many things.

At least I need such a list, up to you if we want to maintain this in the docs this way, or if we would need a dynamic way...?!

### DIFF
--- a/Documentation/Developer/InterlinkInventories.rst
+++ b/Documentation/Developer/InterlinkInventories.rst
@@ -38,7 +38,7 @@ This would output:
 Available default inventories
 -----------------------------
 
-These inventories can be used in any rendered documentation:
+These inventories can be used by default in any rendered documentation:
 
 
 *   Title: :doc:`changelog:Index`
--- a/Documentation/Developer/InterlinkInventories.rst
+++ b/Documentation/Developer/InterlinkInventories.rst
@@ -35,7 +35,7 @@ This would output:
 
 ..  _available-default-inventories:
 
-Currently available default inventories
+Available default inventories
 ---------------------------------------
 
 These inventories can be used in any rendered documentation:
--- a/Documentation/Developer/InterlinkInventories.rst
+++ b/Documentation/Developer/InterlinkInventories.rst
@@ -33,6 +33,135 @@ This would output:
 *   :ref:`TYPO3 Explained 12 LTS <t3coreapi/v12:start>`
 *   :ref:`TYPO3 Explained 12.4 <t3coreapi/12.4:start>`
 
+..  _available-default-inventories:
+
+Currently available default inventories
+---------------------------------------
+
+These inventories can be used in any rendered documentation:
+
+
+*   Title: :doc:`changelog:Index`
+
+    Inventory key: :doc:`changelog <changelog:Index>`
+
+    URL: https://docs.typo3.org/c/typo3/cms-core/main/en-us/
+
+*   Title: :doc:`t3coreapi:Index`
+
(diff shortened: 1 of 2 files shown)


Documentation/Developer/InterlinkInventories.rst
```diff
@@ -33,6 +33,135 @@ This would output:
 *   :ref:`TYPO3 Explained 12 LTS <t3coreapi/v12:start>`
 *   :ref:`TYPO3 Explained 12.4 <t3coreapi/12.4:start>`
 
+..  _available-default-inventories:
+
+Currently available default inventories
```
brotkrueml, line 38: I would avoid "currently". Either it is "current" or "outdated". Additionally, when is "currently"?

garvinhicking: "Currently" is an adverb modifying the adjective "available," indicating that the inventories are available at the present time. Therefore, "currently available" functions as a compound adjective modifying "inventories." So "currently" is "at the time you are reading this".

garvinhicking: Ah, I thought you wanted to write "Current available default inventories". I'm fine with just writng "Available default inventories", that removes redundancy, ack!

garvinhicking:
```suggestion
Available default inventories
```

Documentation/Developer/InterlinkInventories.rst
```diff
@@ -33,6 +33,135 @@ This would output:
 *   :ref:`TYPO3 Explained 12 LTS <t3coreapi/v12:start>`
 *   :ref:`TYPO3 Explained 12.4 <t3coreapi/12.4:start>`
 
+..  _available-default-inventories:
+
+Available default inventories
+---------------------------------------
```
garvinhicking, line 39:
```suggestion
-----------------------------
```

Documentation/Developer/InterlinkInventories.rst
```diff
@@ -33,6 +33,135 @@ This would output:
 *   :ref:`TYPO3 Explained 12 LTS <t3coreapi/v12:start>`
 *   :ref:`TYPO3 Explained 12.4 <t3coreapi/12.4:start>`
 
+..  _available-default-inventories:
+
+Available default inventories
+-----------------------------
+
+These inventories can be used in any rendered documentation:
```
brotkrueml, line 41: Maybe: "... can be used by default in any rendered ..."

This makes it clearer that there is no manual configuration needed.

garvinhicking, line 41:
```suggestion
These inventories can be used by default in any rendered documentation:
```

garvinhicking: Ack, good idea!

brotkrueml (approved):
Fine for me ...

linawolf:
I feel like this is something, that we should rather put in how to document?

linawolf (approved):

garvinhicking:
I'll merge it to this repo first. In "How to document" it would be more of a hassle to maintain across repositories and with different releases. I'd rather link to the docs from "How to document" to here then.

PR #536, opened 2024-05-11
[BUGFIX] Make confval tables responsive

linawolf:


linawolf:
Preventing them from making the mobile layout break by adding a scrollbar instead

PR #565, opened 2024-05-20
[FEATURE] Add back to list / back to top buttons to confvals

linawolf:
image

### DIFF
--- a/tests/Integration/tests/confval/confval-menu-tree/input/index.rst
+++ b/tests/Integration/tests/confval/confval-menu-tree/input/index.rst
@@ -6,11 +6,14 @@ Confval menu as list
 ====================
 
 ..  confval-menu::
+    :name: confval-list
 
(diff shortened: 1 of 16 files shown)


tests/Integration/tests/confval/confval-menu-tree/input/index.rst
```diff
@@ -6,11 +6,14 @@ Confval menu as list
...
 ..  confval-menu::
+    :name: confval-list
 
 Confval menu as list with more info
 ===================================
 
 ..  confval-menu::
+    :caption: Confval menu as list with more info
+    :name: confval-list-info
     :type:
```
froemken, line 17: Do we need type and default here?

linawolf: Yes, they indicate that type and default values from the listed confvals should be displayed

PR #573, opened 2024-05-22
Clarify project

garvinhicking:
People may think that "/project" is a directory they need to replace with the name of their project. Clarify this.

(Open for different wording)

Note: The README.rst.txt does not handle "include::" so instead I opted for a inline code note about "/project".


linawolf (approved):
Thanks a lot!

PR #578, opened 2024-05-27
[BUGFIX] Restore synchronisation of tabs sharing the same name

### LINKED ISSUE #353: Clicking on one tab should also switch to the same tab of every tab group
In the Sphinx-based rendering, when I switch to a tab, all other tabs of every tab group switch to this tab (if available).

This is especially useful, for example, on the commands page: when switching to "Legacy installation", the tab is switched for every group. So the user does not have to switch manually every tab group.

With PHP-based rendering the other tab groups remain on the first tab as active.

mouflondarko:
Resolves #353


linawolf (approved):

linawolf:
Thanks, works great for me!

PR #579, opened 2024-05-27
[TASK] Unify the appearance of horizontal buttons

### LINKED ISSUE #474: Style horizontal buttons
Rendertest has this definition:

https://github.com/TYPO3-Documentation/render-guides/blob/main/Documentation-rendertest/ListsAsButtons/Index.rst

When rendered via `make rendertest`, you can see the output in `Documentation-GENERATED-rendertest/ListsAsButtons/Index.html`, which currently looks like this:

Screenshot 2024-03-28 at 15 39 59
Screenshot 2024-03-28 at 15 40 09
Screenshot 2024-03-28 at 15 40 17

This has a few issues:

* It is too colorful
* It has no WCAG AAA rating

So we need to change all the buttons to match a better scheme, also one where we use much less color to convey information. The distinction between all the classes isn't too great, but many docs have used one button or the other.

So we thought to make different attributions look similar, because their intent is similar.

This is what we came up with:

image

For the warnings we have too ideas; one uses a AA instead of AAA full red display, the other just a border and black text that has AAA then.

If the color scheme like that is ok we can make a PR based on it, and we would also adjust the other buttons we have (search button in the top right, and we'll check for others).

mouflondarko:
Change search button variant to light
Change github button to dark background
Change horizbuttons to use only two colorschemes
Add different hover states to buttons

Resolves #474
___
1. search, view source, how to edit, github buttons
BEFORE:
Zrzut ekranu 2024-05-27 o 15 51 19
AFTER:
Zrzut ekranu 2024-05-27 o 15 50 41

2. Secondary buttons hover state (same state applies to "Search", "Edit on Github" and all "dark" horizontal buttons)
BEFORE:
Zrzut ekranu 2024-05-27 o 15 54 28
AFTER:
Zrzut ekranu 2024-05-27 o 15 55 08

3. Horizontal buttons (attention|important|primary|striking|warning use gray background with white text, note|tip light gray background with gray text)
BEFORE:
Zrzut ekranu 2024-05-27 o 15 53 21
AFTER:
Zrzut ekranu 2024-05-27 o 15 52 43
Dark horizontal button hover state (2nd from left):
Zrzut ekranu 2024-05-27 o 15 57 54

### DIFF
--- a/packages/typo3-docs-theme/resources/public/css/theme.css
+++ b/packages/typo3-docs-theme/resources/public/css/theme.css
@@ -24805,6 +24805,7 @@ ul[class*=horizbuttons-] > li {
 ul[class*=horizbuttons-] > li a {
   color: inherit;
   font-weight: bold;
+  text-decoration: none;
 }
 ul[class*=horizbuttons-] > li p {
   display: inline;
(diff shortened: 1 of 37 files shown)


linawolf:
Thanks a lot, I love It!

garvinhicking:
(Just a post-merge observation: Changes were made in the .css file but we need to do that in both .css AND the .scss file, otherwise generating the assets will always overwrite the .css file)

PR #582, opened 2024-05-28
[BUGFIX] Add proper styling in scss for horizontal buttons

mouflondarko:
This MR will fix the missing scss statements related to horizontal buttons styling changes introduced in #579


linawolf (approved):
Thanks a lot for the follow up

Issue #358, opened 2024-02-21
Languages in version selector not sorted alphabetically


mbrodala:
The version selector includes language variants of the docs. These are currently not sorted by any apparent order:

![image](https://github.com/TYPO3-Documentation/render-guides/assets/5037116/825c8469-2183-4b82-b38c-12abe0fc9952)

These should at least be sorted alphabetically, so that e.g. `12.4 en-us` comes before `12.4 ru-ru`.

linawolf:
@mbrodala if we sort them alphabetically, de-DE  would come before en-us. or should we always put  english first I wonder?

mbrodala:
If there ever is `de-de` then that would be consistent indeed. But at this point the languages should be extracted from the version selector and put somewhere else.

brotkrueml:
For the records: It is not part of the render-guides, but of the PHP script output that is requested by JavaScript to display it on the docs server via Ajax. The Sphinx-based rendering uses the same script.

linawolf:
We should be able to interpret the data provided by the Ajax in different ways though?

linawolf:
We could also adjust the Ajax response here: https://github.com/TYPO3GmbH/site-intercept/tree/develop/legacy_hook

brotkrueml:
> We should be able to interpret the data provided by the Ajax in different ways though?

AFAIR the response is pure HTML, no structured data. This is not a job in JavaScript afterwards.

garvinhicking:
Closing this; the PR #583 has implemented a better sorting and output for this. If there are any objections or improvements, please feel free to create a follow-up issue! :-)

mbrodala:
Not sure why we need another issue given that this one is not fixed yet:

![image](https://github.com/TYPO3-Documentation/render-guides/assets/5037116/7fa19bad-b7bd-4a70-90f8-f47d7f98ce74)

linawolf:
We already deployed a solution which is applied on the next rendering of a manual. You can see it tested out here: https://docs.typo3.org/m/typo3/tutorial-getting-started/draft/en-us/

garvinhicking:
@mbrodala It will only be fixed once a documentation is rendered anew. This does not happen automatically because docs are only rendered when they get new versions.

We'll probably re-render ALL "main" documentations at some point when the PHP-based rendering is stabilized thoroughly.

I'll let @linawolf decide when to close this issue then, because you're probably right that you want to see it before we actually close the issue ;)

[...]

PR #583, opened 2024-05-28
[TASK] Improve AJAX version switcher to sort and group languages and links, document workflow

garvinhicking:
# WHY

Currently the AJAX version selector to change the context of a rendered documentation to another language or version has multiple issues:

* It cannot be used locally without errors
* It cannot be tested locally to simulate the "deployed" behaviour
* It uses pre-rendered HTML from the server
* The language sorting is uninuitive 
* the language naming with ISO-codes is not userfriendly (see Issue #358 )

This PR hopefully 'catches them all'.

# CAVEAT

One extra problem is important to note: I initially wanted to adjust this server-side. However, the code used for that (https://github.com/garvinhicking/site-intercept/commit/aec6c1e3f5bd4dabf949e9bccb4b1eeea6767128) is in the domain of the TYPO3 GmbH, and the Documentation team cannot easily take over Maintenance for an active server-side component. Thus it was decided to fix the problem client-side, using the provided JSON data from the docs.typo3.org API (https://github.com/TYPO3GmbH/site-intercept/pull/158).

# WHAT

This PR contains:

* Updated JavaScript, which now fetches JSON instead of HTML. And parses the JSON, and converts it using sorting methods and static translation name keys to a dynamic HTML list. The HTML structure can be adjusted in the future. 
* The "In one file" section is now a HTML5 'details' element that can be toggled
* The sorting is now:
  * English first: 
    * "main" for english
    * all numbered versions for english (recent version first)
    * all "fixed name" versions for english (alphabetically descending)
  * All other languages sorted alphabetically, using the same sub-ordering than for english
* A PHP proxy is delivered, so that when using a local webserver to view the rendered HTML you can bypass CORS and use the docs.typo3.org API to simulate rendering a version selection menu.
* The HTML DOM is adjusted for new data-attributes so that menu simulation can be performed
* (Hopefully) exhaustive documentation for this is provided with this PR in `Documentation/Developer/AjaxVersions.rst` with more insight

The most glaring sad trombone sound stems from a necessary static language fixation. The `versions.js` file needs to be updated for new languages published to docs.typo3.org. On the upside, for years now the docs were only rendered in english, and only 2 or 3 documentations exist in another language. For a future improvement maybe we find a JSON language file that has a list of language keys like 'de-de' and their associated language name, that we can use or enhance. It's not so often that new languages get added.

# HOW

Details are, as mentioned, in `Documentation/Developer/AjaxVersions.rst`. To quickly simulate this you must:

* Check out this PR
* Start ddev if you have not already
* `make docs`
* Copy the file `packages/typo3-docs-theme/assets/js/versions-proxy.php` to `Documentation-GENERATED-temp/versions-proxy.php`
* `ddev launch` to view the rendered Documentation
* Click on the version selector menu
* Enjoy the sample

You probably want to simulate a different menu getting rendered. You can do this by:

* Reload the HTML URL (you can only do this once per page load, once the AJAX has been called, it will not be called again)
* Open browser JS console, enter: `document.getElementById('toc-version').setAttribute('data-override-url-self', 'https://docs.typo3.org/other/t3docs/render-guides/main/en-us/Developer/InterlinkInventories.html');`
* Click on the version selector menu
* See the selector rendering the output of `render-guides`.

# OUTLOOK

We can possibly improve the design of the language selector, and adjust the nesting appearance for languages. Maybe we can also separate the language selection from the version selection.


linawolf:
Great work, thank you! I made some Small adjustments (label, assets regeneration)

linawolf (approved):

PR #584, opened 2024-05-28
[TASK] Improve nested links readability in main menu

### LINKED ISSUE #562: Improve menues with many levels and longer titles
For exampe https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Backend/JavaScript/RequireJS/Extensions/Index.html

image

Compare with the old menu (we dont want orange back for accessibility reasons) but there was more space and it looked more tidy:
https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/

image

mouflondarko:
Resolves #562

Reduced the font-size of links in menu, changed vertical lines so that they appear only on first nesting level and reduced left offset for nested links

BEFORE:
Zrzut ekranu 2024-05-28 o 13 44 52

AFTER:
Zrzut ekranu 2024-05-28 o 13 43 49

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/layout/navigation/_main_navigation.scss
+++ b/packages/typo3-docs-theme/assets/sass/layout/navigation/_main_navigation.scss
@@ -38,25 +38,32 @@
         padding: calc(#{$spacer} / 4) 0;
         padding-right: 1.5em;
         text-decoration: none;
-	&:hover {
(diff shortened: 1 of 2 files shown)


linawolf (approved):
I think this is an improvement and a good intermediate solution

Issue #229, opened 2023-12-23
Overhaul menu HTML structure


linawolf:

tmotyl:
issues: 
- simplify html and css structure for the menu
- improve the accessibility for the menu (navigation with keyboard)

PR #586, opened 2024-05-28
[TASK] (WIP) Limit the visible links in menu to currently browsing tree

mouflondarko:
Relates to #229


linawolf:
We will not continute in this direction as it is probably to confusing

PR #589, opened 2024-05-29
[BUGFIX] Fixes version switch sorting for versions ending in ".0"

### LINKED ISSUE #588: Versionswitch: versions like 4.0 get converted to 4
This is also messing up the sorting, see https://docs.typo3.org/p/georgringer/news/draft/en-us/

garvinhicking:
Closes #588


packages/typo3-docs-theme/assets/js/versions.js
```diff
@@ -137,7 +137,10 @@
         let versionTrimmed = version.trim();
         let versionAsFloat = parseFloat(versionTrimmed);
         if (!isNaN(versionAsFloat) && Number(versionTrimmed) === versionAsFloat) {
-          version = versionAsFloat;
+          version = versionAsFloat.toString();
```
linawolf, line 140: What would happen to version 4.10 here? I think it would be reduced to 4.1

garvinhicking: Yes, that happens.

I yield for now.

garvinhicking: I found this in a StackOverflow, need to verify:

```
function parseAndPreserveFloat(str) {
    // Parse the float value from the string
    let num = parseFloat(str);
    
    // Check if the original string contains ".0"
    if (str.includes('.') && str.endsWith('.0')) {
        return num.toFixed(1); // Ensures one decimal place
    }
    
    return num.toString(); // Otherwise, return the float as a string
}
```

mouflondarko:
Added a fix to sorting - based on this solution: https://stackoverflow.com/a/40201629
My previous attempt was to include a raw string straight from the AJAX call, although it wasn't enough for versions like `4.9 -> 4.10` - this neat solution from SO ensures that string-based sorting works as expected, and it's reversible when rendering

linawolf (approved):

PR #667, opened 2024-07-25
[TASK] Clarify TextRole popup titles

garvinhicking:
As noted in https://github.com/TYPO3-Documentation/DocsTypo3Org-Homepage/issues/243 this uses a more verbose
tooltip title for specifically highlighted code, see:

Screenshot 2024-07-25 at 12 33 31

### DIFF
--- a/packages/typo3-docs-theme/src/TextRoles/FluidTextTextRole.php
+++ b/packages/typo3-docs-theme/src/TextRoles/FluidTextTextRole.php
@@ -21,6 +21,6 @@ final class FluidTextTextRole implements TextRole
 
     public function processNode(DocumentParserContext $documentParserContext, string $role, string $content, string $rawContent): InlineNode
     {
-        return new CodeInlineNode($rawContent, 'Fluid', 'Templating engine used by TYPO3.');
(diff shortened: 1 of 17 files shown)


linawolf (approved):

linawolf:
Please have a look at the failing tests

garvinhicking:
Probably because integration test files were not updated. Can't take care of this, probably only after my vacation after devdays.

brotkrueml (commented):
Some nit-picking: The "Code written in PHP" does not apply to FQCN like \TYPO3\CMS\Backend\Controller\SomeController, previous "PHP" was more generic and valid for the FQCN.

fe-hicking:
What would you propose? "Code with PHP Syntax"?

brotkrueml:
> What would you propose? "Code with PHP Syntax"?

To be really precise, a new role like "php-class" or "php-fqcn" would be appropriate which then may be titled "PHP class name". Already said, just nit-picking, no blocking. Feel free to merge.

linawolf:
We could also auto-detect FQNs, we already have a regex for that

linawolf:
Lets merge it like this now, we can always refine it later.

PR #680, opened 2024-07-29
[FEATURE] Help texts for inline PHP code snippets

linawolf:
image
image
image
image
image

Short view is opt-in and can for example be used if the same class is mentioned several times in the same context:

```
:php-short:`\TYPO3\CMS\Adminpanel\ModuleApi\ContentProviderInterface``
``

### DIFF
--- a/packages/typo3-docs-theme/src/TextRoles/PhpTextRole.php
+++ b/packages/typo3-docs-theme/src/TextRoles/PhpTextRole.php
@@ -5,10 +5,19 @@ namespace T3Docs\Typo3DocsTheme\TextRoles;
 use phpDocumentor\Guides\Nodes\Inline\InlineNode;
 use phpDocumentor\Guides\RestructuredText\Parser\DocumentParserContext;
 use phpDocumentor\Guides\RestructuredText\TextRoles\TextRole;
+use T3Docs\GuidesPhpDomain\PhpDomain\FullyQualifiedNameService;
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: packages/typo3-docs-theme/src/TextRoles/PhpTextRole.php [hunk 1]
```diff
@@ -21,6 +26,48 @@ public function getAliases(): array
...
+                '
+            );
+        }
+        if (str_starts_with($rawContent, '$GLOBALS[\'TSFE\']')) {
+            return new CodeInlineNode(
+                $rawContent,
+                'TypoScriptFrontendController',
+                'TSFE is short for \TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController,
+                    a class which exists in the system extension EXT:frontend.
+                    It is available as global array $GLOBALS[\'TSFE\'] in PHP.
```
brotkrueml, line 67: It is also available as PSR-7 request attribute frontend.controller in frontend (which should be preferred instead of the global variable) - but only for newer versions. I assume, detecting the version is not possible?

linawolf: We will output this help text, wherever a string in `:php:` tag starts with $GLOBALS['TSFE']. So when the global variable is not mentioned anymore, the help text will not be displayed anymore. Or only in the changelog for old versions.

PR #681, opened 2024-07-29
[TASK] Change color and thickness of table of contents title

### LINKED ISSUE #625: Style the project title
Project titles can be at times quite long. They are then in several lines and not good readable. Also the orange is not accessible

For example:
https://docs.typo3.org/m/typo3/guide-contributionworkflow/main/en-us/
image

https://docs.typo3.org/other/typo3/view-helper-reference/12.4/en-us/
image

Find an accessible and easy to read way to display these titles.

mouflondarko:
Resolves #625 
___

Replace orange color with dark gray, reduce font-size from 1.5em to 1.1em and increase font-weight from 600 to 700
Examples with longer title as provided in the source task:

BEFORE:
image

AFTER:
image


linawolf:
I wonder if we could still use some kind of Orange highlight, not for the text color but some element?

mouflondarko:
@linawolf maybe some kind of orange bottom border below version switch/headline itself, it could introduce a bit of color here (although it would be purely decorational, so I'm not sure)
Orange underline text-decoration for headline could also be considered, but it gets noisy:
image

linawolf:
@mouflondarko Maybe an orange vertical border to the left?

mouflondarko:
@linawolf It does look a bit better, but we sacrifice some space we gained from reducing the font-size: 
image

as an alternative - left border without affecting the box itself, nonobstructive but it's leaving the standard viewport:
image

linawolf:
I like the second option better. It would be good to align it with navigation headers in manuals that have those:
image
Note: I did not manage to do this without including the version switch but you probably can to better then me ;)

See for example https://docs.typo3.org/m/typo3/guide-contributionworkflow/main/en-us/WorkflowExplained/Index.html

linawolf:
Me too, @mouflondarko  can you make us an orange border version?

mouflondarko:
I've made the changes on this branch, now the headline looks like this:
image

For reference, longer headline looks like this:
image

In order to make this change, I moved the entire header slightly to the right (0,5em) in order to match the left border with navigation headers and keep the spacing between border and the headline

linawolf (approved):

PR #701, opened 2024-08-13
[TASK] Refactor site settings directive

linawolf:


froemken (approved):
I like the new structure, the checks before using the array and all the clean castings before using the variable. Good work. Not tested in real-life, but code looks pretty good.

PR #702, opened 2024-08-14
[FEATURE] Allow "EXT:self" syntax for typo3:site-set-settings

garvinhicking:
By default only files within a "Documentation/" directory can be read by the "typo3:site-set-settings" directive.

However in cases where extensions (Core or third party) provide additional files in their main directory, these should be accessible, to allow reading the ACTUAL file contents, and not requiring a copy to be placed inside Documentation/.

With this patch, a new syntax "EXT:self" allows to refer to this base "/project/" directory as a path prefix, instead of "/project/Documentation/". Other file traversal beyond this is forbidden.

### DIFF
--- a/packages/typo3-docs-theme/src/Directives/SiteSetSettingsDirective.php
+++ b/packages/typo3-docs-theme/src/Directives/SiteSetSettingsDirective.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace T3Docs\Typo3DocsTheme\Directives;
 
+use League\Flysystem\Adapter\Local;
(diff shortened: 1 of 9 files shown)


linawolf:
Very cool! I think it should be possible to test this out with the examples extension and a local docker container?

garvinhicking:
I tested this already with your core patch to the fluid_styled_content extension, and also with a custom extension of mine.

The next step would be IMO to merge it so it can be tested in GitHub Action workflows...?

linawolf (approved):

linawolf:
Ok merged it and will create a new tag now

PR #718, opened 2024-08-26
[BUGFIX] Fix word wrapping for long confval titles

### LINKED ISSUE #698: [BUG] Long confval titles do not break in mobile view
image

They should break where there is a non alphanumeric sign like a dot or dash. If they don't have any they should just break anywhere.

mouflondarko:
Fixes #698
This rather simple fix makes confval titles break preferably by white space, if there is no white space present or word is too long it will break anywhere.

BEFORE:
obraz

AFTER:
obraz

Note: Non-alphanumeric signs like dots or dashes are not taken into account when wrapping - in order to break titles by them, it would be necessary to parse the title beforehand to insert the [Line Break Opportunity `<wbr>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr) when needed (e.g. before a dot or a dash), like so:

```html
this.text.breaks.anywhere
this<wbr>.text<wbr>.breaks<wbr>.on<wbr>.dots
```

Here is the preview for titles using the `<wbr>` tags (inserted manually):
```html
email<wbr>.templateRootPaths
expose<wbr>Nonexistent<wbr>User<wbr>In<wbr>Forgot<wbr>Password<wbr>Dialog
```
obraz

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/components/directives/_confval.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/directives/_confval.scss
@@ -6,6 +6,8 @@ dl.confval {
     padding-bottom: $spacer * .3;
     border: solid 3px $light;
     border-top-color: $gray-500;
+    word-wrap: anywhere;
+    white-space: normal;
     & > dt {
         display: block;
         background-color: $light;
--- a/packages/typo3-docs-theme/assets/sass/components/directives/_confval.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/directives/_confval.scss
@@ -15,6 +15,8 @@ dl.confval {
         margin-bottom: .75em;
         code {
             color: color-contrast($light);
+            word-wrap: anywhere;
+            white-space: normal;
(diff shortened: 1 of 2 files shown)


linawolf:
While you are at it, could you make sure that the other texts like the default value in the screenshot are also breaking?

garvinhicking:
@linawolf could we use this <wbr> and insert it automatically for siteset directives and others too, basically after any "." character?

Copying such words should not copy hidden chars then since it's only invisible html markup....?

Sounds great to me :)

linawolf:
@garvinhicking we could make a twig function for that I guess

mouflondarko:
@linawolf sure, I've pushed a commit that resolves the issue (the <wbr> tags would also work for those values):
obraz

linawolf (approved):

linawolf:
Thanks for this change, merged, I will collect several changes before I make a new release however.

Could you open an Issue about the wbr tags?

PR #735, opened 2024-09-05
[TASK] Apply updated CGL rules due to #731

garvinhicking:


garvinhicking:
Merging as low-brainer.

PR #734, opened 2024-09-05
[TASK] Allow settings labels to be provided in xlf files

bnf:
Required prerequisite for:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/85869
https://forge.typo3.org/issues/104831

PR #737, opened 2024-09-06
[TASK] Support auto-registration of set labels from labels.xlf

bnf:
The set language files introduced in https://forge.typo3.org/issues/104831 are now auto-loaded from labels.xlf within the sets folder, if the file is provided and the set configuration does not reference an explicit labels xlf file.

### DIFF
--- a/packages/typo3-docs-theme/src/Directives/SiteSetSettingsDirective.php
+++ b/packages/typo3-docs-theme/src/Directives/SiteSetSettingsDirective.php
@@ -83,15 +83,16 @@ final class SiteSetSettingsDirective extends BaseDirective
         }
 
         $labelContents = '';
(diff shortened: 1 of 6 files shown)


bnf:
The xlf handling introduced yesterday – https://github.com/TYPO3-Documentation/render-guides/pull/734 – is to be slightly extended to support autoloading.
Could you please review  & release, so that the following patch can be merged (I don't want to break documentation rendering, so I'm waiting until this is merged)
> 85907: [TASK] Auto-register set labels from labels.xlf | https://review.typo3.org/c/Packages/TYPO3.CMS/+/85907

linawolf (approved):

garvinhicking:
Thank you for addressing it :)

PR #742, opened 2024-09-11
[TASK]: Add all-documentations-menu WebComponent

### LINKED ISSUE #566: Develop a WebComponent for a central documentation menu

mouflondarko:
Closes #566 

For presentation purposes, all data is mocked as a static JS object


linawolf:
@mouflondarko thanks a lot for this solution! We like it very much.

I rendered a JSON containting all the manuals here: https://docs.typo3.org/h/typo3/docs-homepage/main/en-us/mainmenu.json 

Is it possible to use the result of that JSON to render the menu?

mouflondarko:
@linawolf sure, the structure looks good and should work as expected. Should the WebComponent fetch this JSON client-side?

mouflondarko:
@linawolf I've introduced the change in fetching data for the menu, now it should pull current configuration from the url you provided
I noticed a CORS-related issue was already encountered in this project (for version switcher), so I did basically the same (tried to modify the original file to make "universal" proxy, but it wasn't working properly, probably due to special characters in URLs – because of limited time I just copied the existing proxy file and slightly modified it, so for local preview a file `menu-proxy.php` needs to be copied to `Documentation-GENERATED-temp/`

With the current configuration menu looks like this:
obraz
Not sure about this long column for System Extensions - I could try splitting long columns

linawolf:
I can also split the output of the System extensions in two for the JSON. that is probably faster then a CSS / JavaScript solution

linawolf:
See https://github.com/TYPO3-Documentation/DocsTypo3Org-Homepage/pull/259, the columns should look better now

mouflondarko:
@linawolf I think it looks good 👍 
obraz

linawolf:
I think so too, what else is needed to make it ready for review?

mouflondarko:
@linawolf I think it's ready, would appreciate a feedback

linawolf (approved):

linawolf:
Looks good, I will merge, make a release and try it out in the wild

PR #765, opened 2024-09-24
[TASK] Add "--minimal-test" CLI option

### LINKED ISSUE #764: Add "--minimal-test" option
See https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/4777:

With an input option like '--minimal-test' we can control performance optimizations.

garvinhicking:
This allows us to control specific settings needed for mininmal tests (like setting only a single output format) or more for the future.

Closes #764

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,12 +124,12 @@ test: test-integration test-unit test-xml test-docs test-rendertest ## Runs all 
 .PHONY: test-docs
 test-docs: ## Runs project generation tests
 	@echo "$(ENV_INFO)"
-	$(PHP_BIN) vendor/bin/guides --no-progress Documentation --output="/tmp/test" --config=Documentation --fail-on-log
+	$(PHP_BIN) vendor/bin/guides --no-progress Documentation --output="/tmp/test" --config=Documentation --minimal-test
 
 .PHONY: test-rendertest
 test-rendertest: ## Runs rendering with Documentation-rendertest
 	@echo "$(ENV_INFO)"
-	$(PHP_BIN) vendor/bin/guides --no-progress Documentation-rendertest --output="Documentation-GENERATED-rendertest" --config=Documentation-rendertest --fail-on-log
+	$(PHP_BIN) vendor/bin/guides --no-progress Documentation-rendertest --output="Documentation-GENERATED-rendertest" --config=Documentation-rendertest --minimal-test
 
 .PHONY: rendertest
 rendertest: ## Runs rendering with Documentation-rendertest
--- a/packages/typo3-docs-theme/src/EventListeners/AddThemeSettingsToProjectNode.php
+++ b/packages/typo3-docs-theme/src/EventListeners/AddThemeSettingsToProjectNode.php
@@ -17,6 +17,21 @@ final class AddThemeSettingsToProjectNode
     public function __invoke(PostProjectNodeCreated $event): void
     {
         $projectNode = $event->getProjectNode();
+
+        // Native parsing of argv because we do not have the original ArgvInput
+        // available, and neither the InputDefinition. That's ok for the
+        // very basic parsing of a global option.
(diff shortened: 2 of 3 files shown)


linawolf:
I was testing this out both with 
```
docker run --rm -v ${PWD}:/project -it typo3-docs:local --progress --config=Documentation --minimal-test
```
and ddev
```
ddev exec vendor/bin/guides Documentation --config=Documentation --minimal-test
```` 
In both cases all 3 versions got rendered:

```
docker run --rm -v ${PWD}:/project -it typo3-docs:local --progress --config=Documentation --minimal-test
Parsing: 13/13 [============================] 100% Parsed 13 files in 0.67 seconds
Rendering: 13/13 [============================] 100% Output format html: Rendered 13 documents in 2.99 seconds                                                                                                                      
Rendering: 13/13 [============================] 100% Output format singlepage: Rendered 13 documents in 0.03 seconds                                                                                                                
Rendering: 13/13 [============================] 100% Output format interlink: Rendered 13 documents in 0.00 seconds
Successfully placed 13 rendered HTML, SINGLEPAGE, and INTERLINK files into /project/Documentation-GENERATED-temp
```

linawolf:
This might be due to the fact, that the main guides.xml also defines the output formats and these probably overrule the command? Not sure

garvinhicking:
You are probably right. I tested it without the 'config' directive. Need to inspect.

garvinhicking:
OK, this is too complex, I sadly have to give up. The guides.xml is loaded by guides after our decorator, there are dozens of config loaders and processors involved and I could not find a possibility to hook into it with for example an event listener, where I also would be able to get the original CLI input arguments.

We might need to first add this to `guides` to be able to override guides.xml options. I've asked Jaap and begged for help :D

garvinhicking:
@linawolf Jaap hinted me at the PostProjectNodeCreated event in which I can hook into. It's not very beautiful code because I natively parse argv here, because the event doesn't seem to offer access to the original InputInterface from Symfony nor the InputDefinition.

Since we only access that single "--minimal-test" option I don't feel too bad about it.

However, I'm unsure what the 'fail-on-log' config option would do. I only can access 'setFailOnError()' on the projects settings, to 'Fail-on-log' seems to be something else. Do we need it? If so, we would probably need to set that one in the RunDecorator, but if that option is then part of a 'guides.xml' we cannot override it.

linawolf:
I am not really sure what fail on log did

garvinhicking:
Then I would suggest to continue ignoring it for the time being. The important flags (fail on error, singlepage output format) should work with the updated patch.

linawolf:
Did you test if we also fail on warning?

garvinhicking:
I only tested if the option was set and implied it should work. Needs testing/verification, you are right.

linawolf:
I just tested this and adjusted the make files to use the flag. How ever, once I use `--minimal-test` no matter if I additionally --fail-on-log or not the runs are not returning a non-null status

linawolf:
I build one warning into the docs so we can test this

garvinhicking:
@linawolf Hm, the return status should be unaffected by this?! And "fail-on-log" is also unaffected. The only thing the patch at this point should do is adjust the output format, and enforce "fail-on-error" (NOT on warning).

I think warnings already have a success return code. Only errors would have another return code. Is the thing you added an error or a warning?

linawolf:
The thing I added is a warning. you can try this out by calling "make test-docs" locally. If you you remove "--minimal-test" from the command in the makefile the command fails due to the warning an the "fail on log". If you add "--minimal-test" it does not fail.

garvinhicking:
Alright, found it, I think.

'--fail-on-error' sets 'failOnError=error'
'--fail-on-log' sets 'failOnError=warning'

Thus I changed the logic for the 'minimal-test' case to set failonError('warning').

linawolf (approved):

PR #780, opened 2024-10-10
[BUGFIX]: Fix all-documentations-menu versions, positioning and UX on mobile devices

mouflondarko:
1. **Fix versions not being displayed in the menu**
   - BEFORE:
  obraz
   - AFTER:
  obraz

2. **Fix stretched-link anchor tags being rendered over the menu**
     Due to implementation of Bootstrap's `stretched-link` class, anchor tags were rendered over all-documentations-menu popover window, making them clickable through the menu's background. Managed to fix this by unsetting the z-index of such anchor tags which got rid of the problem, and it seems that no cards were negatively impacted by this change (user is still able to click anywhere within the card)

3. **Fix visible height on mobile devices**
    On mobile devices the menu was partially covered by the floating button. In order to fix this, I've reduced the max height of the popover window and made it scrollable on mobile, now the menu shouldn't be covered by the button and I think this change is good for UX on mobile, as now user is certain that he is scrolling through a menu (previously to see all links within the menu, the whole page would be scrolled)
   - BEFORE:
   obraz
   - AFTER (notice the scrollbar thumb on the right hand-side of popver menu):
   obraz


mouflondarko:
4. Changed the incorrect "All documentations" (plural) button label to "All documentation" (singular)

linawolf (approved):

PR #781, opened 2024-10-10
[FEATURE] Add "Shortlink" feature to docs.typo3.org/permalink/

garvinhicking:
With https://github.com/TYPO3GmbH/site-intercept/pull/188 we introduced the possibility on docs.typo3.org to forward to interlink shortcodes to the appropriate targets.

This server-side script is able to go through all
`objects.inv.json` files hosted on the server to parse the proper targets, and then do a HTTP forward to
the actual documentation files.

This patch implements a new entry block "Shortlink" (because "permalink" is the actual link to the site, and still has value to be displayed as a primary
entry) in the popup modal when clicking on a header reference.

Screenshot 2024-10-10 at 20 47 26

NOTE: This should be regarded as an "experimental" feature. There may be situations in which the
resolution of permalink targets with confvals and
filenames and others may not work properly yet. We will need to catch these issues and address
them with follow-up patches to either this
repository or the site-intercept repository
using specific examples.


garvinhicking:
**Please note, this has no priority**

garvinhicking:
```
make integration-baseline
```

seems to currently mangle:

```
tests/Integration/tests-full/main-menu-json/expected/mainmenu.json
```

Please advise :)

linawolf:
@garvinhicking after rebasing onto main this does not happen anymore.

linawolf (approved):

bnf:
The shortlink for https://docs.typo3.org/c/typo3/cms-sys-note/main/en-us/Editor/Index.html#for-editors
is generated as

```
https://docs.typo3.org/permalink/typo3/cms-sys-note:for-editors
```

Which results in
```
Invalid shortcode.
```

I first thought that might be caused by underscore vs dash issues, but EXT:core seems to be affected as well:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/Howto.html

Although the shortlink looks "more" broken (missing `typo3/cms-core` reference?):
```
https://docs.typo3.org/permalink/changelog:Changelog/Howto#documenting-changes
```


Which also results in
```
Invalid shortcode.
```

garvinhicking:
@bnf Thanks. Created issue https://github.com/TYPO3-Documentation/render-guides/issues/812 - will look into it. Some headings, dependant on their creation method, might be stored dofferently.

PR #788, opened 2024-10-16
[FEAT] use current filename to get the original file name

jaapio:
This will make markdown rendering work with less errors.


jaapio:
Waiting for a new guides release :-)

PR #796, opened 2024-10-21
[TASK] Allow issue links to bitbucket

linawolf:


packages/typo3-docs-theme/src/Twig/TwigExtension.php
```diff
@@ -342,8 +342,16 @@ public function getReportIssueLink(array $context): string
         if (str_starts_with($reportButton, 'https://gitlab.com/')) {
             return $reportButton;
         }
+        if (str_starts_with($reportButton, 'https://bitbucket.org/')) {
+            $reportButton = $this->enrichBitbuckedReport($reportButton, $renderContext);
+            return $reportButton;
+        }
         if ($reportButton !== '') {
-            $this->logger->warning('For security reasons only only "report-issue" links in the guides.xml to a local page (starting with "/") or to one of these 3 plattforms are allowed: https://forge.typo3.org/ https://github.com/ https://gitlab.com/');
+            $this->logger->warning(
```
liayn, line 350: maybe make this a notice/info only?
Warnings are usually something, which a an author wants to get rid of, which is not possible in that case. => notice/info

linawolf: If you set the issue link but it is not being displayed that is worth a warning. You can get rid of the warning by removing the link as it is not displayed anyway?

liayn: Okay, I misunderstood this. I thought the link is always displayed somewhere, but only the button is not rendered. If this is not the case, then of course a warning is okay.

PR #804, opened 2024-10-30
559 style command lists

Tim-Obert:
resolves #599 

image

the commands are now styled like the confvals. 

Also I added the watch command in the Gruntfile.js

And I fixed the offset of the copy button in scrolling code blocks: 
Before:
image

After: 
image


Tim-Obert:
Closed because of merge issues. Cherrypicked in #809

PR #817, opened 2024-11-07
Fix link reference copy button + Check if Modals exist

Tim-Obert:
Fix link reference-shortlink copy button. 
The JS led to errors because the Modals don't exist eg. in https://github.com/TYPO3-Documentation/t3docs-search-indexer


garvinhicking:
Thank you Tim for fixing my bug with the copy-button misreference, much appreciated.

Rest looks good to me, but only checked superficially.

garvinhicking (approved):

PR #837, opened 2024-11-20
[FEATURE] Make convert markdown to reST available

linawolf:
You can try out conversion with
```make mdtest```


garvinhicking (approved):
Hahah wow, this feels like magic :D

PR #848, opened 2024-11-23
[FEATURE] Make migration of inventories possible

linawolf:
I had to drop support for non-stable prefered TYPO3 versions in migration. That would mean in such a case they are listed as single <inventroy> tags


garvinhicking (approved):
I think this should work and prove to be very helpful!

PR #883, opened 2025-01-06
[TASK] Follow-up: CDN change to cdn.typo3.com

garvinhicking:
References #880


linawolf (approved):
Thanks for catching this!

PR #926, opened 2025-03-15
[FEATURE] Replace Tooltip with Modal for Code and Class Information

### LINKED ISSUE #743: Use popover for PHP class name explanation
Currently hovering a PHP class name with the `php` role yields a tooltip with a short explanation:

image

However, it would be better if this explanation was triggered via click/keyboard instead. For example using the `(?)` on the right as trigger.

Benefits:

1. Better accessibility by enabling keyboard access
2. Option to move the API/docs link into the popover; this allows for easier copy and paste of PHP class names without unintentionally opening a link.
3. Less "noise" in general since the explanation is really only shown on intent, never by accident.

linawolf:
Replaced the tooltip with a modal to provide a more detailed and interactive interface for displaying code and class information.

image

Resolves #743


mbrodala:
Lovely, thanks!

PR #1033, opened 2025-08-24
[BUGFIX] Typo inside the aria role of the all documentation menu

ineswillenbrock:
Possible typo inside the aria role of the all documentation menu


linawolf:
@jaapio Do you have any idea why the validate monorepo structure fails? It should have no relation to this PR

jaapio:
No I think the tool has been updated. Or some other package that now contains a version that's not compatible. we do install the tool, and there is no lock file.

PR #1054, opened 2025-09-12
[TASK] Overwrite run command

jaapio:
Other than what we did before we do now a full overwrite of the run command. This allows us to remove a number of less optimal constructions in the command. The change itself is empowered by the changes in phpdocumentor/guides-cli 1.9 which now provides a number of helper classes to simplify self build run commands.


linawolf (approved):
Looks good to me, we should definately try this out!

garvinhicking:
We will hold off merging this for a short time to first have a stable release, and then integrate this :)

PR #1073, opened 2025-10-01
[TASK] Fix local rendering in ddev

ineswillenbrock:
* independecy of the port of the local instance


linawolf (approved):

linawolf:
Thanks for taking care of this!

PR #1091, opened 2025-10-26
[TASK] Speed up menu rendering

jaapio:
In the original template all files in the documentation contained the full menu structure. While not all was ever used. By rendering just the sub menu's when they are active we do improve the speed of rendering drastically. This results in a **80%** performance boost for projects like changelog.


linawolf (changes requested):
Unfortunarily this would break the menu where you can use the toggle buttons to go to other parts of the docs.

If we put this behind a --fast option or some such it might help local rendering however

CybotTM:
Hi @jaapio,

maybe I missed something or did not get the full context, as it seems a lot of intrinsic knowledge is not documented here in the PR,  but

IMO this PR is superseded by patches provided at deeper layers: AST caching, template caching, O(1) document lookups, and parallel processing. In upstream phpDocumentor/guides:

- phpdocumentor/guides#1287
- phpdocumentor/guides#1288
- phpdocumentor/guides#1289
- phpdocumentor/guides#1293
- phpdocumentor/guides#1294
- phpdocumentor/guides#1295
- phpdocumentor/guides#1296
- phpdocumentor/guides#1297

This is also where other projects will benefit from it.

So I will focus elsewhere.

linawolf:
The solution doesnt work as is, we might start a new attempt in the future

PR #1117, opened 2025-12-08
[BUGFIX] Improve error handling for invalid guides.xml configuration

CybotTM:
## Summary

Instead of crashing with a PHP Fatal error when guides.xml contains invalid configuration, the guides entrypoint now catches `InvalidTypeException` and `InvalidConfigurationException` and displays a helpful error message.

**Before:**
```
PHP Fatal error:  Uncaught Symfony\Component\Config\Definition\Exception\InvalidTypeException: 
Invalid type for path "guides.theme". Expected "scalar", but got "array". in ...
Stack trace:
#0 ...
(20+ lines of stack trace)
```

**After:**
```
 Invalid guides.xml configuration 

Your guides.xml file contains a configuration error:
  Invalid type for path "guides.theme". Expected "scalar", but got "array".

Common causes:
  - Invalid or unknown XML element
  - Missing required attributes
  - Invalid attribute values or types

Run vendor/bin/typo3-guides lint-guides-xml for detailed validation.
See: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/GeneralConventions/GuideXml.html
```

## Approach

This is a **minimal fix** that catches exceptions and provides a user-friendly error message. It's simple (~25 lines) but doesn't provide line numbers or specific validation details.

## Alternative

See also the companion PR with a **comprehensive fix** that validates against XSD schema before loading, providing exact line/column numbers: #1118

## Test plan

- [ ] Create a malformed guides.xml with `<theme name="typo3docs" />` as child element
- [ ] Run `bin/guides run --config=Documentation Documentation`
- [ ] Verify helpful error message is shown instead of PHP Fatal error
- [ ] Verify valid guides.xml files still render correctly

### DIFF
--- a/tests/Integration/InvalidGuidesXmlTest.php
+++ b/tests/Integration/InvalidGuidesXmlTest.php
@@ -22,7 +22,7 @@ final class InvalidGuidesXmlTest extends TestCase
     protected function setUp(): void
     {
         $this->tempDir = sys_get_temp_dir() . '/render-guides-invalid-test-' . uniqid();
-        mkdir($this->tempDir, 0755, true);
+        mkdir($this->tempDir, 0o755, true);
 
         // Copy fixture files to temp directory
         copy(self::FIXTURE_SOURCE . '/guides.xml.fixture', $this->tempDir . '/guides.xml');


garvinhicking (approved):
I think this looks helpful... not too sure about possible impacts when catching these very specific exceptions, so I'd like some more vote/feedback on this patch, especially due to the AI usage.

PR #1118, opened 2025-12-08
[BUGFIX] Validate guides.xml against XSD schema before loading

CybotTM:
## Summary

Builds on the runtime exception handler from #1117 by adding upfront XSD schema validation of `guides.xml` files. When the XML is malformed, users now see the exact line and column of each problem instead of the generic message produced by the catch fallback.

**Before this PR (post-#1117):**

```
 Invalid guides.xml configuration

Your guides.xml file contains a configuration error:
  Invalid type for path "guides.theme". Expected "scalar", but got "array".
...
```

**After:**

```
 Invalid guides.xml configuration

Your guides.xml file failed XSD schema validation:

Schema validation failed for Documentation/guides.xml
  Line 3, Column 0: Element '{https://www.phpdoc.org/guides}theme': This element is not expected.

See: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/GeneralConventions/GuideXml.html
```

## How it works

- Resolves the project- and command-level `guides.xml` paths the same way the loader does
- Validates each one against `vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd` via `DOMDocument::schemaValidate`
- On failure, prints file path + line/column for every libxml error and exits before the Symfony loader runs
- The `InvalidTypeException` / `InvalidConfigurationException` catch added in #1117 is kept as a safety net for anything XSD can't express

## Test plan

- [ ] Malformed `guides.xml` with `<theme name="typo3docs" />` as a child element → exact line/column shown
- [ ] Valid `guides.xml` files render unchanged


linawolf:
@CybotTM is this change still needed or is the small version we already merged sufficient?

CybotTM:
Hi @linawolf 

#1117 did only properly catch the error. It will produce a generic error.

#1118 does a complete schema validation which should produce a more helpful error message.

(Description updated)

linawolf (approved):

PR #1139, opened 2025-12-22
feat(figure): add image zoom functionality with multiple modes

CybotTM:
## Summary

Add `:zoom:` option to figure and image directives supporting four zoom modes:

- **lightbox**: Click to open full-size in modal dialog overlay
- **gallery**: Click to open with navigation between grouped images  
- **inline**: Mouse wheel zoom directly on image with pan support
- **lens**: Magnifying lens follows cursor showing zoomed view

## Usage

```rst
.. figure:: /images/screenshot.png
   :zoom: lightbox

   Click to enlarge this screenshot

.. figure:: /images/photo1.png
   :zoom: gallery
   :gallery: photos

   Part of the photos gallery
```

## Features

- `:zoom:` option accepts `lightbox` | `gallery` | `inline` | `lens`
- `:gallery:` option groups images for gallery navigation
- `:zoom-indicator:` option to show/hide the zoom icon (default: show)
- Zoom indicator with appropriate cursor (`zoom-in` or `crosshair`)
- Tooltips on indicators ("Click to enlarge", "Scroll to zoom")
- Click/scroll events pass through indicator to underlying image
- Full keyboard accessibility for all zoom modes
- Respects `prefers-reduced-motion` media query
- Works with `with-border` and `with-shadow` styling classes

## Technical Changes

- New `FigureDirective.php` extending `SubDirective`
- New `image.html.twig` template for standalone images with zoom
- JavaScript in `image-zoom.js` bundled into `theme.min.js`
- CSS in `_component_image_zoom.scss` imported into `theme.scss`
- Updated `figure.html.twig` with zoom data attributes

## Test plan

- [ ] Verify lightbox mode opens dialog on click
- [ ] Verify gallery mode with navigation between grouped images
- [ ] Verify inline mode with scroll-to-zoom and drag-to-pan
- [ ] Verify lens mode with magnifying glass following cursor
- [ ] Verify keyboard navigation works for all modes
- [ ] Verify zoom indicator positioning with border/shadow classes
- [ ] Verify `:zoom-indicator: false` hides the indicator

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3243,16 +3243,16 @@
         },
         {
             "name": "symfony/console",
-            "version": "v6.4.30",
+            "version": "v6.4.31",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/console.git",
-                "reference": "1b2813049506b39eb3d7e64aff033fd5ca26c97e"
+                "reference": "f9f8a889f54c264f9abac3fc0f7a371ffca51997"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/console/zipball/1b2813049506b39eb3d7e64aff033fd5ca26c97e",
-                "reference": "1b2813049506b39eb3d7e64aff033fd5ca26c97e",
+                "url": "https://api.github.com/repos/symfony/console/zipball/f9f8a889f54c264f9abac3fc0f7a371ffca51997",
+                "reference": "f9f8a889f54c264f9abac3fc0f7a371ffca51997",
                 "shasum": ""
             },
             "require": {
@@ -3317,7 +3317,7 @@
                 "terminal"
             ],
             "support": {
-                "source": "https://github.com/symfony/console/tree/v6.4.30"
+                "source": "https://github.com/symfony/console/tree/v6.4.31"
             },
             "funding": [
                 {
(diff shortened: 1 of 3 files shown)


linawolf:
Happy to see you are working on this, please ping me when you require my review

josefglatz:
I just tested it as a non-a11y-expert. Really cool to see them accessible.

is it also possibel with actual RsT parser to have a automatic rendered thumbnail/image in the normal view and an original image in the lightbox?

CybotTM:
@josefglatz Good question! Currently the zoom modes use the same image source for both the inline view and the lightbox/gallery.

To support separate thumbnail/original images, we'd need to add a new directive option like `:zoom-src:` that specifies the full-size image to show in the lightbox:

```rst
.. figure:: /images/screenshot-thumbnail.png
   :zoom: lightbox
   :zoom-src: /images/screenshot-original.png
   :alt: Screenshot

   Click to see full resolution
```

This would be a nice enhancement for a follow-up PR. Would you like to create a feature request issue for it?

linawolf:
I would suggest that we first get this PR merged and then deal with different images for preview and the zoom box itself in follow ups. 

I am not very familiar with how the rendering pipeline actually works so I would be hesistant to change it unless someone really knows what they are doing. New features are nice, but it also means that we have more things we need to maintain and promise for the indefinate future. Also ppl actually using such features are usually quite a small group. So I would prefer a lowtech variant like  :zoom-src: /images/screenshot-original.png In such a case we should also link the original picture so it is accessible without javascript

linawolf (approved):
Thanks a lot for these changes

PR #1143, opened 2025-12-29
[FEATURE] PHP 8.5 with parallel rendering and performance optimizations

CybotTM:
## Summary

Major performance optimization introducing parallel rendering and incremental builds. Achieves **10x faster** renders on multi-core systems.

**📊 [View Performance Report](https://cybottm.github.io/render-guides/)**

### Performance Results

| Scenario | Main Branch | Feature Branch | Improvement |
|----------|-------------|----------------|-------------|
| Cold Build (38 docs) | 27.2s | 2.7s | **10x faster** |
| Warm Build (no changes) | 27.1s | 0.6s | **45x faster** |
| Partial Build (1 file changed) | 27.1s | 0.8s | **34x faster** |

### Key Features

- **Parallel Rendering**: Fork-based parallel compilation and rendering using `pcntl_fork`
- **Incremental Builds**: Skip unchanged documents based on content hashing and dependency tracking
- **AST Caching**: Cache parsed document ASTs to avoid re-parsing unchanged files
- **Twig Template Caching**: Pre-compile Twig templates for faster rendering
- **Inventory Caching**: Cache interlink inventory JSON to avoid repeated HTTP requests
- **OPcache CLI**: Enable OPcache in Docker CLI for improved PHP performance
- **Dependency Graph**: Track cross-document dependencies (toctree, includes, references)
- **Global Invalidation Detection**: Detect changes requiring full re-render

### Build Metadata

Stores `_build_meta.json` in output directory containing:
- Content hashes (xxh128)
- Export data (anchors, section titles, citations)
- Dependency graph
- Settings hash for invalidation detection

## Delivery Plan

This PR serves as an **exploratory/WIP branch**. The optimizations are being delivered incrementally:

### Phase 1: Caching Infrastructure (Upstream)

| PR | Description | Status |
|----|-------------|--------|
| [#1287](https://github.com/phpDocumentor/guides/pull/1287) | Rendering caching layer | ✅ All CI pass |
| [#1288](https://github.com/phpDocumentor/guides/pull/1288) | RST parsing optimizations | ✅ All CI pass |
| [#1289](https://github.com/phpDocumentor/guides/pull/1289) | CLI container caching | ✅ All CI pass |
| [#1291](https://github.com/phpDocumentor/guides/pull/1291) | Symfony 8 compatibility | ✅ Merged |
| [#1293](https://github.com/phpDocumentor/guides/pull/1293) | O(1) document entry lookups | ✅ Merged |
[...]

PR #1144, opened 2026-01-04
[BUGFIX] Fix broken confval permalink reference in test documentation

CybotTM:
## Summary

Fixes broken permalink reference in `Documentation-rendertest/Confval/Index.rst` that causes test failures with `--minimal-test` flag.

## Discovery

While working on PR #1143 (performance optimizations), CI tests failed with:
```
Reference confval-case-array could not be resolved in Confval/Index
```

This was causing `make test-rendertest` to exit with code 1 because `--minimal-test` sets `failOnError('warning')`.

## Technical Analysis

The broken line was:
```rst
Permalink to confval: `array of cObjects  <https://docs.typo3.org/permalink/rendertest:confval-case-array>`_.
```

### Why it fails

The permalink URL `rendertest:confval-case-array` is processed by `ReplacePermalinksNodeTransformer` which creates:
- `ReferenceNode` with `targetReference='confval-case-array'`
- `linkType='std:label'` (default)

But the actual confval anchor (defined in `ConfvalTrees.rst` with `:name: case-array`) is registered as:
- `anchor='case-array'` (without the `confval-` prefix)
- `linkType='std:confval'`

The reference resolver looks for:
```php
$internalLinkTargets['std:label']['confval-case-array']
```

But the target exists at:
```php
$internalLinkTargets['std:confval']['case-array']
```

**They don't match!** The permalink format doesn't support specifying link types.

### The correct approach

Using the proper `:confval:` text role:
```rst
Permalink to confval: :confval:`case-array` (defined in ConfvalTrees.rst).
```

This creates a `ReferenceNode` with `linkType='std:confval'` and `anchor='case-array'`, which resolves correctly.

## Question for maintainers

This broken reference was introduced in PR #975 (commit c310707a) on May 4, 2025 by @linawolf.

**Was this intentional?** Possible interpretations:
1. **Typo/mistake**: The author expected `confval-case-array` to work like `:confval:`case-array`` but they're fundamentally different
2. **Intentional test case**: Perhaps this was meant to test error handling for broken permalinks?
3. **Missing feature**: Should the permalink format support typed references like `rendertest:confval:case-array`?

If this was intentional as a "broken link" test case, please let me know and I'll adjust the fix accordingly (perhaps adding a comment explaining it's intentionally broken, or moving the test).

## Test plan

- [ ] Verify `make test-rendertest` passes
- [ ] Verify the confval link renders correctly in generated HTML

### DIFF
--- a/Documentation-rendertest/Confval/Index.rst
+++ b/Documentation-rendertest/Confval/Index.rst
@@ -6,7 +6,7 @@
 confval
 =======
 
-Permalink to confval: `array of cObjects  <https://docs.typo3.org/permalink/rendertest:confval-case-array>`_.
+Permalink to confval: :confval:`case-array` (defined in ConfvalTrees.rst).
 
 ..  toctree::
     :glob:


CybotTM:
Closing this PR - the original permalink format was correct and should work.

## Root Cause Found

The actual bug was a **priority ordering conflict** in the feature branch's incremental rendering infrastructure:

| Component | Priority | Function |
|-----------|----------|----------|
| `CollectLinkTargetsTransformer` | 5000 | Registers primary link targets |
| `ExportsCollectorPass` | 4500 | Collects internal targets for incremental cache |
| `CollectPrefixLinkTargetsTransformer` | **4000** (was) | Registers `std:label` entries for prefixed nodes like `confval-*` |

Since higher priority runs first, `ExportsCollectorPass` was collecting targets **before** `CollectPrefixLinkTargetsTransformer` added the `confval-*` entries to `std:label`. This caused the permalink `rendertest:confval-case-array` to fail resolution.

## Fix Applied

Changed `CollectPrefixLinkTargetsTransformer.getPriority()` from `4000` to `4900` in commit https://github.com/CybotTM/render-guides/commit/d4c3eeb9.

This ensures it runs **before** `ExportsCollectorPass` (4500) but **after** `CollectLinkTargetsTransformer` (5000).

## Why Only Feature Branch?

`ExportsCollectorPass` is new incremental rendering infrastructure only present on the feature branch. Main doesn't have this component, so there was no priority conflict - explaining why main never showed this issue.

The original documentation was correct - no changes needed there.

PR #1179, opened 2026-02-17
[FEATURE] Modernize float classes to Bootstrap 5 with deprecation handling

CybotTM:
## Summary

Builds on #1174 (merged). Modernizes legacy float CSS classes to Bootstrap 5 logical properties with backward-compatible deprecation handling.

- Modernize CSS float selectors to Bootstrap 5 logical properties (`float-start`/`float-end`), keeping backward-compatible aliases with grouped selectors to reduce duplication
- Add Twig `alignMap` to translate RST `:align:` values to CSS classes, with deduplication to prevent duplicate classes when both `:align:` and `:class:` resolve to the same value
- Detect and rewrite deprecated `:class: float-left`/`:class: float-right` in both FigureDirective and ImageDirective, emitting deprecation warnings
- Extract shared `RewritesLegacyFloatClasses` trait for detection/rewrite logic (DRY)
- ImageDirective uses **composition** (decorator pattern) instead of forking the upstream `final` class — zero code duplication, upstream changes are inherited automatically
- Add responsive breakpoint for standalone images on small screens

## Implementation notes

- **ImageDirective**: Wraps the upstream `final` class via composition. Intercepts `process()` to rewrite float classes, then delegates to `$this->inner->process()`. No need to duplicate `processNode()` or `resolveLinkTarget()`.
- **FigureDirective**: Existing fork extended with float class detection. Uses `array_filter`/`explode` to strip float classes from the inner `` while preserving non-float classes (e.g. `with-shadow`). A future PR will explore refactoring this to use decoration as well.
- **RewritesLegacyFloatClasses trait**: Shared regex detection (`hasLegacyFloatClass`) and rewrite (`rewriteLegacyFloatClasses`) used by both directives.
- Upstream issue for `final` removal: https://github.com/phpDocumentor/guides/issues/1303

## Test plan

- [x] Integration tests cover: align left/right/center, legacy float-left/right rewriting, mixed classes, simultaneous :align: + :class:, standalone images, modern float-start/float-end pass-through, substitution images
- [x] Deprecation warnings verified for all legacy class usages (7 warnings, no false positives for modern classes)
- [x] Unit tests: 40 tests covering RewritesLegacyFloatClasses trait, ImageDirective, and FigureDirective (100% patch coverage)

### DIFF
--- a/Documentation-rendertest/ImagesAndFigures/FloatAndAlignment.rst
+++ b/Documentation-rendertest/ImagesAndFigures/FloatAndAlignment.rst
@@ -20,7 +20,7 @@ Image float left
 
 .. |example-teaser-left| image:: ../images/q150_cccccc.png
    :alt: Left floating image
-   :class: float-left with-shadow
+   :class: float-start with-shadow
 
 |example-teaser-left|
 Typesetting is the composition of text by means of arranging physical
@@ -41,7 +41,7 @@ Image float right
 
 .. |example-teaser-right| image:: ../images/q150_cccccc.png
    :alt: Right floating image
-   :class: float-right with-shadow
+   :class: float-end with-shadow
 
 |example-teaser-right|
 Typesetting is the composition of text by means of arranging physical
@@ -62,7 +62,8 @@ Figure float left
 
 .. figure:: ../images/q150_cccccc.png
    :alt: Left floating figure
-   :class: float-left with-shadow
+   :align: left
+   :class: with-shadow
 
    A figure floated to the left
 
@@ -85,7 +86,8 @@ Figure float right
 
 .. figure:: ../images/q150_cccccc.png
    :alt: Right floating figure
-   :class: float-right with-shadow
+   :align: right
+   :class: with-shadow
 
    A figure floated to the right
 
--- a/packages/typo3-docs-theme/assets/sass/components/_images.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_images.scss
@@ -17,15 +17,20 @@ figure {
     &:not(:first-child) {
         @extend .mt-3;
     }
+    &.float-start,
+    &.float-end,
+    // Deprecated: legacy Bootstrap 4 class names — remove after migration period (see #1179)
     &.float-left,
     &.float-right {
         margin-bottom: $spacer;
         max-width: 50%;
     }
+    &.float-start,
     &.float-left {
         float: left;
         margin-right: $spacer;
     }
+    &.float-end,
     &.float-right {
         float: right;
         margin-left: $spacer;
@@ -37,6 +42,8 @@ figure {
 }
 
 @media (max-width: 575.98px) {
+    figure.float-start,
+    figure.float-end,
     figure.float-left,
     figure.float-right {
         float: none;
--- a/packages/typo3-docs-theme/assets/sass/components/_images.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_images.scss
@@ -21,17 +21,22 @@ figure {
         @extend .mt-3;
     }
 
+    &.float-start,
+    &.float-end,
+    // Deprecated: legacy Bootstrap 4 class names — remove after migration period (see #1179)
     &.float-left,
     &.float-right {
         margin-bottom: $spacer;
         max-width: 50%;
     }
 
+    &.float-start,
     &.float-left {
         float: left;
         margin-right: $spacer;
     }
 
+    &.float-end,
     &.float-right {
         float: right;
         margin-left: $spacer;
@@ -44,6 +49,8 @@ figure {
 }
 @media (max-width: 575.98px) {
 
+    figure.float-start,
+    figure.float-end,
     figure.float-left,
     figure.float-right {
         float: none;
--- a/packages/typo3-docs-theme/assets/sass/components/_rst.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_rst.scss
@@ -32,11 +32,14 @@
         display: inline-block;
         max-width: 100%;
         height: auto;
+        // Deprecated: legacy Bootstrap 4 class names — remove after migration period (see #1179)
+        &.float-start,
         &.float-left {
             float: left;
             margin-right: $spacer;
             margin-bottom: $spacer;
         }
+        &.float-end,
         &.float-right {
             float: right;
             margin-left: $spacer;
@@ -55,6 +58,17 @@
                 margin-right: 0;
                 max-width: 100%;
             }
+        }
+    }
+    @media (max-width: 575.98px) {
+        img.float-start,
+        img.float-end,
+        img.float-left,
+        img.float-right {
+            float: none;
+            max-width: 100%;
+            margin-left: 0;
+            margin-right: 0;
         }
     }
     .plantuml object {
--- a/packages/typo3-docs-theme/assets/sass/components/_rst.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_rst.scss
@@ -35,12 +35,15 @@
         max-width: 100%;
         height: auto;
 
+        // Deprecated: legacy Bootstrap 4 class names — remove after migration period (see #1179)
+        &.float-start,
         &.float-left {
             float: left;
             margin-right: $spacer;
             margin-bottom: $spacer;
         }
 
+        &.float-end,
         &.float-right {
             float: right;
             margin-left: $spacer;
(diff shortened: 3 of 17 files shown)


CybotTM:
@ineswillenbrock 

> Did you want to rewrite the deprecated float-left/right to float-start/end or just log their usage?

Both.

linawolf:
Thanks for the rebase — I checked this out locally against current `main`: it merges cleanly, and the full suite (`make test`, `make code-style`, `make phpstan`) is green, so nothing else is holding it up.

Two things I'd like fixed before we merge, though. Both are small.

**1. Bootstrap's float utilities override the new mobile resets**

Bootstrap 5 ships `.float-start` / `.float-end` as utilities, and they carry `!important`:

```css
/* resources/public/css/theme.css:19094 */
.float-start { float: left !important; }
.float-end   { float: right !important; }
```

There is no `.float-left` / `.float-right` utility in Bootstrap 5 — which is exactly why the theme's own float rules were unopposed until now.

After the rename, the new breakpoint blocks in `_images.scss:50` and `_rst.scss:70` declare a plain `float: none`, and that loses to `!important` no matter how specific the selector is. Since `figure.html.twig` now maps `:align: left` to `float-start` rather than `float-left`, this is a behaviour change for documents that already exist: a `.. figure:: :align: left` un-floats correctly below 576px on `main`, but on this branch it keeps floating at `max-width: 100%`, so the figure takes the full width and the following text tries to wrap beside it.

Adding `!important` to the `float: none` in both breakpoint blocks (and regenerating `theme.css`) should be all it needs.

**2. The rewrite regex matches hyphenated project classes**

In `RewritesLegacyFloatClasses.php:23`, `\b` treats `-` as a word boundary, so project-specific class names match too:

```php
preg_replace('/\bfloat-left\b/', 'float-start', 'my-float-left');
// => 'my-float-start'
```

That rewrites an author's own class out from under their CSS and emits a deprecation warning that doesn't apply to it. It matters more here than before, because the previous regex only stripped classes on figures, rather than rewriting and warning on both directives. Anchoring on class boundaries would fix it:

```php
'/(?<![\w-])float-(left|right)(?![\w-])/'
```

The rest of what I noticed is cosmetic and can wait for a follow-up. Happy to push these two onto the branch myself if that's easier — "Allow edits from maintainers" is on.

linawolf:
@CybotTM I've looked at the two points from my review again, and found a third gap in the same place. Here's a patch with all three. It applies to the current head of your branch (55d617c36) with `git apply`:

1. The mobile `float: none` resets in `_images.scss` and `_rst.scss` now carry `!important`, so they win against Bootstrap's `.float-start`/`.float-end` utilities (both are `!important`; the theme's selectors win on specificity).
2. The `object` reset in `_rst.scss` only listed `float-left`/`float-right`; it now covers `float-start`/`float-end` too.
3. `RewritesLegacyFloatClasses` is anchored on class boundaries (`(?<![\w-])float-(left|right)(?![\w-])`), so a project class like `my-float-left` is left alone. There are unit tests for the hyphenated cases.

The patch leaves out the compiled `theme.css`. Could you rebase onto current `main` and then run `make assets`? The branch is 25 commits behind, several of which change the theme's SCSS and JS, so the compiled files need to be built from the combined sources. The build also runs `stylelint --fix`, which adds two blank lines in `_rst.scss` in your first commit; the patch already contains them.

I tried it that way locally: rebased onto `main`, patch applied, assets rebuilt, and `make pre-commit-test` is green. After that, it's ready to merge from my side.

<details>
<summary>Patch</summary>

```diff
diff --git a/packages/typo3-docs-theme/assets/sass/components/_images.scss b/packages/typo3-docs-theme/assets/sass/components/_images.scss
index 95074ae34..1e0dd6716 100644
--- a/packages/typo3-docs-theme/assets/sass/components/_images.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_images.scss
@@ -53,7 +53,8 @@ figure {
     figure.float-end,
     figure.float-left,
     figure.float-right {
-        float: none;
+        // !important: Bootstrap's .float-start/.float-end utilities carry it
+        float: none !important;
         max-width: 100%;
         margin-left: 0;
         margin-right: 0;
diff --git a/packages/typo3-docs-theme/assets/sass/components/_rst.scss b/packages/typo3-docs-theme/assets/sass/components/_rst.scss
index f0b065770..ff422b406 100644
--- a/packages/typo3-docs-theme/assets/sass/components/_rst.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_rst.scss
@@ -36,6 +36,7 @@
         height: auto;
 
         // Deprecated: legacy Bootstrap 4 class names — remove after migration period (see #1179)
+
         &.float-start,
         &.float-left {
             float: left;
@@ -57,22 +58,27 @@
         }
         @media (max-width: 575.98px) {
 
+            // !important: Bootstrap's .float-start/.float-end utilities carry it
+
+            &.float-start,
+            &.float-end,
             &.float-left,
             &.float-right {
-                float: none;
+                float: none !important;
                 margin-left: 0;
                 margin-right: 0;
                 max-width: 100%;
             }
         }
     }
-
     @media (max-width: 575.98px) {
+
         img.float-start,
         img.float-end,
         img.float-left,
         img.float-right {
-            float: none;
+            // !important: Bootstrap's .float-start/.float-end utilities carry it
+            float: none !important;
             max-width: 100%;
             margin-left: 0;
             margin-right: 0;
diff --git a/packages/typo3-docs-theme/src/Directives/RewritesLegacyFloatClasses.php b/packages/typo3-docs-theme/src/Directives/RewritesLegacyFloatClasses.php
index 14f32a7dc..937943a2b 100644
--- a/packages/typo3-docs-theme/src/Directives/RewritesLegacyFloatClasses.php
+++ b/packages/typo3-docs-theme/src/Directives/RewritesLegacyFloatClasses.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace T3Docs\Typo3DocsTheme\Directives;
 
 use function preg_match;
-use function preg_replace;
+use function preg_replace_callback;
 
 /**
  * Shared logic for detecting and rewriting deprecated Bootstrap 4 float class names.
@@ -18,16 +18,22 @@ use function preg_replace;
  */
 trait RewritesLegacyFloatClasses
 {
+    /**
+     * Anchored on class boundaries rather than \b, which also breaks at a
+     * hyphen: a project class such as "my-float-left" is not Bootstrap's.
+     */
+    private const LEGACY_FLOAT_CLASS_PATTERN = '/(?<![\w-])float-(left|right)(?![\w-])/';
+
     private function hasLegacyFloatClass(string $classValue): bool
     {
-        return (bool) preg_match('/\bfloat-(left|right)\b/', $classValue);
+        return (bool) preg_match(self::LEGACY_FLOAT_CLASS_PATTERN, $classValue);
     }
 
     private function rewriteLegacyFloatClasses(string $classValue): string
     {
-        return (string) preg_replace(
-            ['/\bfloat-left\b/', '/\bfloat-right\b/'],
-            ['float-start', 'float-end'],
+        return (string) preg_replace_callback(
+            self::LEGACY_FLOAT_CLASS_PATTERN,
+            static fn(array $matches): string => $matches[1] === 'left' ? 'float-start' : 'float-end',
             $classValue,
         );
     }
diff --git a/packages/typo3-docs-theme/tests/unit/Directives/RewritesLegacyFloatClassesTest.php b/packages/typo3-docs-theme/tests/unit/Directives/RewritesLegacyFloatClassesTest.php
index 811b041b5..4f9374a97 100644
--- a/packages/typo3-docs-theme/tests/unit/Directives/RewritesLegacyFloatClassesTest.php
+++ b/packages/typo3-docs-theme/tests/unit/Directives/RewritesLegacyFloatClassesTest.php
@@ -52,6 +52,8 @@ final class RewritesLegacyFloatClassesTest extends TestCase
         yield 'empty string' => ['', false];
         yield 'partial match float-leftover' => ['float-leftover', false];
         yield 'partial match afloat-right' => ['afloat-right', false];
+        yield 'hyphenated project class my-float-left' => ['my-float-left', false];
+        yield 'hyphenated project class float-right-wide' => ['with-shadow float-right-wide', false];
     }
 
     #[Test]
@@ -71,5 +73,7 @@ final class RewritesLegacyFloatClassesTest extends TestCase
         yield 'modern classes unchanged' => ['float-start', 'float-start'];
         yield 'no float classes unchanged' => ['with-shadow', 'with-shadow'];
         yield 'empty string unchanged' => ['', ''];
+        yield 'hyphenated project classes unchanged' => ['my-float-left float-right-wide', 'my-float-left float-right-wide'];
+        yield 'legacy class beside a hyphenated project class' => ['my-float-left float-left', 'my-float-left float-start'];
     }
 }
```

</details>

Issue #1188, opened 2026-02-23
[BUG] :rst: text role cannot display a single backslash character


CybotTM:
## Description

The `:rst:` text role (and likely other code-type text roles using `$rawContent`) cannot correctly display special characters that require escaping.

## How to reproduce

Given these RST sources, rendered with render-guides:

```
| Test A (escaped backticks): :rst:`rest \`code\``
| Test B (double backslash): :rst:`\\`
| Test C (plain): :rst:`rest code`
| Test D (double backticks): :rst:`rest ``code```
| Test E (single backslash): :rst:`\`
```

## Rendered results

| Test | RST source | Rendered | Problem |
|------|-----------|-----------|---------|
| A | `:rst:`rest \`code\`` ` | Plain text, role not recognized | Escaped backticks confuse the lexer, role delimiter breaks |
| B | `:rst:`\\`` | `\\` (double backslash displayed) | rawContent preserves literal `\\` instead of resolving to `\` |
| C | `:rst:`rest code`` | `rest code` | OK - works without special characters |
| D | `:rst:`rest ``code``` ` | Broken into `:rst:`rest` + `code`` | Double backticks split the role parsing |
| E | `:rst:`\`` | `:rst:`` ` (empty plain text) | Backslash escapes the closing backtick, role never closes |

**Only completely "harmless" content (Test C) works.** Any content requiring escape sequences fails.

## Root cause

The issue is in the interaction between the RST inline lexer and code-type text roles:

1. The `InlineLexer` processes escape sequences (`\` + char) into `ESCAPED_SIGN` tokens **before** text role content boundaries are determined
2. `RestructuredTextTextRole` (and `LiteralTextRole`) use `$rawContent` per [phpDocumentor/guides#533](https://github.com/phpDocumentor/guides/pull/533), which preserves literal source including escape backslashes
3. This creates an impossible situation: `\\` stays as `\\` in rawContent, while `\` before the closing backtick breaks the delimiter

## Relevant code

- `RestructuredTextTextRole.php` line 30 — uses `$rawContent`
- phpDocumentor/guides `TextRoleRule.php` — builds both `$part` (processed) and `$rawPart` (raw) from `ESCAPED_SIGN` tokens
- phpDocumentor/guides PR https://github.com/phpDocumentor/guides/pull/533 — intentionally made code roles use `rawContent`

## Possible solutions

[...]

PR #1189, opened 2026-02-24
[BUGFIX] Resolve double-backslash escape in code text roles

CybotTM:
## Summary

Demonstrates a possible fix for #1188 — `:rst:` (and all other code-type text roles) display `\\` literally instead of resolving to `\`.

All 20 code-type text roles pass `$rawContent` to `CodeInlineNode`, which preserves RST escape sequences literally. This means `\\` in the RST source stays as `\\` in the rendered output instead of resolving to `\` as the author intended.

**Approach:** Add `str_replace('\\\\', '\\', $value)` in the `CodeInlineNode` constructor to resolve double-backslash escape sequences.

### Changes

- `CodeInlineNode.php` — one line added in constructor
- New integration test `code-textrole-escape` covering `:rst:`, `:shell:`, `:typoscript:`, `:yaml:` with `\\` and plain content
- Render-test examples added to `Inline-code-and-textroles/Index.rst`

### Scope

This PR addresses **Test B** from the issue (double backslash). Tests A, D, E involve the upstream `InlineLexer` in phpDocumentor/guides consuming `` \` `` as `ESCAPED_SIGN` before text role boundaries are resolved — that requires a separate upstream fix.

## ⚠️ Known limitation — this is a demonstration, not a complete fix

The `str_replace` approach is **blind** — it cannot distinguish between:

- `\\` meaning "RST escaped backslash → display `\`" (the bug we're fixing)
- `\\` meaning "literal double backslash in code that should stay as `\\`"

**PHP code samples that would break:**

| RST source | Before this PR | After this PR | Problem |
|---|---|---|---|
| `:php:`'/\\d+/'`` | `'/\\d+/'` | `'/\d+/'` | Lost the PHP string escape |
| `:php:`"hello\\nworld"`` | `"hello\\nworld"` | `"hello\nworld"` | Looks like a newline now |
| `:php:`preg_match('/\\\\/')`` | `preg_match('/\\\\/')` | `preg_match('/\\/')` | Regex completely wrong |

This is the fundamental tension identified in the issue: `$rawContent` was introduced by [phpDocumentor/guides#533](https://github.com/phpDocumentor/guides/pull/533) to preserve literal backslashes for PHP namespaces (`\TYPO3\CMS\Core`), but that same literalness prevents `\\` from resolving to `\`.

**A proper fix needs to happen upstream** in the parser/lexer layer where `ESCAPED_SIGN` tokens are available — only the parser knows which `\\` came from an intentional RST escape vs which `\` was just a literal backslash character in code.

## Test plan

- [x] `make test-integration ENV=local` — all 113 tests pass
- [x] `make code-style ENV=local` — 0 issues
- [x] `make phpstan ENV=local` — no errors
- [ ] `make rendertest ENV=local` — inspect "Escape handling in code text roles" section
- [ ] Verify `:php:` namespace paths still render correctly (existing `code-php` test)

### DIFF
--- a/Documentation-rendertest/Inline-code-and-textroles/Index.rst
+++ b/Documentation-rendertest/Inline-code-and-textroles/Index.rst
@@ -239,6 +239,37 @@ Inline `code` :php:`MyCustomException` :ts:`PAGE` in title
 
 
 
+Escape handling in code text roles
+===================================
+
+Double backslash (``\\``) in code text roles should render as a single
+backslash. This applies to all code-type text roles that use ``$rawContent``.
+
+Source::
+
+   :rst:`a\\b`
+   :shell:`path\\to\\file`
+   :typoscript:`config.no\\cache`
+   :yaml:`key\\value`
+
+Result:
+
+*  :rst:`a\\b`
+*  :shell:`path\\to\\file`
+*  :typoscript:`config.no\\cache`
+*  :yaml:`key\\value`
+
+Plain content without escapes should still work:
+
+*  :rst:`rest code`
+*  :shell:`ls -la`
+
+PHP namespace paths use single backslashes and must be preserved:
+
+*  :php:`\TYPO3\CMS\Core\Utility\GeneralUtility`
+*  :code:`TYPO3\CMS\Core\Cache\Frontend\FrontendInterface`
+
+
 Fully qualified names with backslashes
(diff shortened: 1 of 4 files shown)


CybotTM:
so, yes, should make this "demonstration" obsolete.

PR #1085, opened 2025-10-17
[FEATURE] Make templates overridable

linawolf:
When using the docker container locally
or in a custom pipeline.

PR #1190, opened 2026-02-25
[FEATURE] Make templates overridable

CybotTM:
Lets you override the theme's Twig templates without rebuilding the container, useful for local rendering and custom CI pipelines.

Two override directories are checked, both prepended to the Twig search path so they win over the built-in templates:

1. `/templates` — mounted via Docker volume (highest priority)
2. `resources/custom-templates` — bundled in the project repo

Templates rendered by the central docs.typo3.org pipeline are unaffected: that runs a fixed image and never sees these directories.

Continues #1085 by @linawolf, picking up the path and entrypoint fixes discussed there with @jaapio and @kaystrobach.


CybotTM:
We can do the asset copy in this PR or in a follow-up ... your call @linawolf

kaystrobach:
Thank you

linawolf:
> We can do the asset copy in this PR or in a follow-up ... your call @linawolf

I would suggest to do it in a follow up

linawolf (approved):

PR #1184, opened 2026-02-22
[TASK] Update and SHA-pin all GitHub Actions

CybotTM:
## Summary

Updates and SHA-pins all GitHub Actions in non-CI workflow files for supply
chain security.

`main.yaml` is excluded from this PR — it will be migrated to shared reusable
workflows via [#1196](https://github.com/TYPO3-Documentation/render-guides/pull/1196).

### Files changed

- `.github/workflows/docker.yaml` — metadata-action, login-action, setup-qemu, setup-buildx, build-push, upload/download-artifact, checkout
- `.github/workflows/docker-test.yaml` — checkout
- `.github/workflows/deploy-azure-assets.yaml` — checkout
- `.github/workflows/split-repositories.yaml` — checkout, cache, use-github-token, use-subsplit-publish
- `.github/workflows/pr-auto-merge.yaml` — dependabot/fetch-metadata
- `.github/workflows/pr-auto-approve.yaml` — dependabot/fetch-metadata

### Version updates

All versions verified as latest releases via the GitHub API on 2026-07-04.

| Action | Old | New | SHA |
|--------|-----|-----|-----|
| actions/checkout | SHA only (v6.0.2) | v7.0.0 | `9c091bb2` |
| actions/cache | SHA only (v5.0.3) | v6.1.0 | `55cc8345` |
| docker/metadata-action | SHA only | v6.2.0 | `dc802804` |
| docker/login-action | SHA only | v4.4.0 | `af1e73f9` |
| docker/setup-qemu-action | SHA only | v4.2.0 | `96fe6ef7` |
| docker/setup-buildx-action | SHA only | v4.2.0 | `bb05f3f5` |
| docker/build-push-action | SHA only | v7.3.0 | `53b7df96` |
| actions/upload-artifact | SHA only | v7.0.1 | `043fb46d` |
| actions/download-artifact | SHA only | v8.0.1 | `3e5f45b2` |
| dependabot/fetch-metadata | SHA only | v3.1.0 | `25dd0e34` |
| frankdejonge/use-github-token | SHA only | 1.1.0 | `15e6289d` |
| frankdejonge/use-subsplit-publish | SHA only | 1.1.0 | `00010151` |

### Reviewer note on test coverage

[...]

PR #1196, opened 2026-03-09
[TASK] Migrate CI to shared reusable workflows

CybotTM:
## Summary

Replaces the inline CI jobs in `main.yaml` with the shared workflows of [TYPO3-Documentation/.github](https://github.com/TYPO3-Documentation/.github), so action versions and the composer/cache setup are maintained in one place instead of once per repository.

| Job | Before | After |
|-----|--------|-------|
| Tests | inline matrix | `reusable-php-tests.yml` |
| Quality (CS Fixer, PHPStan, XML lint) | inline | `reusable-php-quality.yml` |
| Composer normalize | step inside Quality | `reusable-php-command.yml` |
| Documentation render test | step inside Quality | `reusable-php-command.yml` |
| Render-test validation | step inside Quality | `reusable-php-command.yml` |
| Validate monorepo structure | inline | `reusable-php-command.yml` |
| typo3-docs-theme JS tests | inline | unchanged — Node, not PHP |

Rewritten against current `main` rather than rebased: the original branch predated 59 commits and would have reverted the `merge_group` trigger, the `test-rendertest` and `test-monorepo` jobs and the theme JS job.

## This needs a ruleset change in the same step

A called workflow reports its checks as `<job name> / <inner job name>`, so six of the seven context names change. These are the names GitHub actually produced on this head, read from the check runs of `39a0585e` — not predicted:

| now required | after this change |
|---|---|
| `Tests (8.2)` … `Tests (8.5)` | `Tests / Tests (PHP 8.2)` … `Tests / Tests (PHP 8.5)` |
| `Quality` | `Quality / Quality` |
| `Validate monorepo structure` | `Validate monorepo structure / Run` |
| `typo3-docs-theme JS tests` | unchanged |
| — | `Composer normalize / Run` |
| — | `Documentation renders without warning / Run` |
| — | `Documentation-rendertest renders without warning / Run` |

The last three deserve attention: `composer normalize --dry-run`, `make test-docs` and `make test-rendertest` used to be *steps inside* the required `Quality` check, so they gated merges implicitly. As separate jobs they are only gated if the ruleset lists them. Leaving them out would weaken the gate as a side effect of the migration, so the recommended required set is all ten contexts above.

Merging this without the ruleset change blocks the merge queue: it would keep requiring contexts that no longer report.

## Order of operations

There is no ordering that keeps the gate closed the whole time. Switching the ruleset first blocks every other open PR, whose checks still report under the old names; merging first blocks the queue. The workable sequence is to drop the required contexts for the merge window and set the new ones immediately after.

## Note

The shared workflows currently pin `actions/checkout` v6.0.2, `shivammathur/setup-php` 2.37.1 and `actions/cache` v5.0.3, while this repository's inline jobs are on newer pins. The migration therefore moves those three back a version; they are maintained centrally by Dependabot in `.github` afterwards.

## Related

- Depends on .github#2 (adds `reusable-php-command.yml`) — merged 2026-03-22.
- SHA-pinning of actions in other workflow files is handled separately in #1184.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,131 +1,56 @@
 name: "Main"
 
-on: # yamllint disable-line rule:truthy
+on:
   pull_request: null
   push:
     branches:
       - "main"
 
-env:
-  DEFAULT_PHP_VERSION: "8.2"
-  RUN_ENVIRONMENT: "local"
+permissions: {}
 
 jobs:
   tests:
-    name : Tests
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        php:
-          - '8.2'
-          - '8.3'
-          - '8.4'
-          - '8.5'
-    steps:
-      - name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
-
-      - name: "Install PHP"
-        uses: shivammathur/setup-php@44454db4f0199b8b9685a5d763dc37cbf79108e1
-        with:
-          coverage: "none"
-          php-version: "${{ matrix.php }}"
-          extensions: 'inotify, pcntl'
-
-      - name: "Get Composer cache directory"
-        id: composer-cache
-        run: echo "dir=$(composer config cache-files-dir)" >> "$GITHUB_OUTPUT"
-
-      - name: "Cache Composer dependencies"
-        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
-        with:
-          path: ${{ steps.composer-cache.outputs.dir }}
-          key: php-${{ matrix.php }}-composer-${{ hashFiles('**/composer.lock') }}
-          restore-keys: php-${{ matrix.php }}-composer-
-
-      - name: "Install dependencies with Composer"
-        run: composer install --no-interaction --no-progress --prefer-dist
-
-      - name: "Run unit tests"
-        run: "make test-unit ENV=${{ env.RUN_ENVIRONMENT }}"
-
-      - name: "Run integration tests"
-        run: "make test-integration ENV=${{ env.RUN_ENVIRONMENT }}"
+    uses: TYPO3-Documentation/.github/.github/workflows/reusable-php-tests.yml@main
+    with:
+      php-versions: '["8.2", "8.3", "8.4", "8.5"]'
+      test-unit-command: "make test-unit ENV=local"
+      test-integration-command: "make test-integration ENV=local"
+      php-extensions: "inotify, pcntl"
 
   quality:
-    name: Quality
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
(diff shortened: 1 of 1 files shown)


ineswillenbrock (approved):
LGTM

CybotTM:
Heads-up from #1326 (merge-queue trial, decision record in TYPO3-Documentation/.github#22): a merge queue on `main` gates on the check contexts reported by this workflow. This PR rewrites the `on:` block and, by moving jobs into reusable workflows, renames the reported contexts (`Tests (8.2)` → nested names).

Two things to coordinate when this PR moves forward:

1. Keep the `merge_group:` trigger from #1326 when rebasing — it lives in the caller workflow.
2. Update the branch ruleset's required status checks to the new context names in the same merge window. With stale context names, every merge group times out and all merging is blocked until the ruleset is fixed.

linawolf:
Thanks for this, and sorry it sat so long — that delay is on us, not on the work.

I'd like to keep render-guides on inline CI for now, so I'm going to close this. To be clear about what that is and isn't: it's not a rejection of shared workflows. The org already uses them widely — `reusable-backport.yml`, `reusable-apply-precommit.yml` and `reusable-test-documentation.yml` are called from a dozen-odd manual repositories, and that's clearly the right shape there.

The PHP workflows are a different case, and the reason is just arithmetic: `reusable-php-tests.yml`, `reusable-php-quality.yml` and `reusable-php-command.yml` have no callers today. The only mention of them anywhere in the org is `.github/README.md`. render-guides would be the sole consumer, so "maintained in one place instead of once per repository" doesn't yet have a second repository to amortise against — we'd be splitting one workflow across two repos and getting the same maintenance back, plus coupling.

Two costs that come with it in the meantime:

**Pinning moves the wrong way.** `main.yaml` SHA-pins every action and Dependabot keeps those current — today they're *ahead* of the shared workflows (`setup-php` 2.37.2 vs 2.37.1, `actions/cache` v6.1.0 vs v5.0.3), as the description notes. And referencing them at `@main` means a push in another repository changes this repository's CI with no review here. For the repo that renders every other manual, I'd rather keep that surface small.

**The ruleset cost is real.** A called workflow reports as `<job> / <inner job>`, so six of seven required contexts rename. The three steps that don't fit `reusable-php-quality.yml` — `composer normalize --dry-run`, `make test-docs`, `make test-rendertest` — become separate jobs that must be listed explicitly or they silently stop gating, having been implicitly gated inside `Quality` until now. With the merge queue from ADR-004 live on `main`, that's a coordination window I don't want to spend for a benefit that isn't there yet.

What would change my mind: a second PHP repository adopting these workflows. At that point the amortisation argument is real, and I'd want render-guides to follow rather than be the holdout. Tag releases with Dependabot bumping consumers, and a generic extra-commands input on `reusable-php-quality.yml`, would each make it easier again.

I'll write this up as a short ADR next to `Documentation/Decisions/ADR-004-MergeQueues.rst` so it's answerable without digging through a closed PR. Separately — we call none of the shared workflows today, and `reusable-apply-precommit.yml` and `reusable-backport.yml` look like they'd fit here. Happy to look at those on their own merits if you're interested.

PR #1237, opened 2026-04-27
[TASK] Drop popper-js & replace with HTML popover, fix & upgrade stylelint

richardkrikler:
Also fixes all-documentations-menu desktop behavior.
Hide below viewport width, as then mobile menu is used.

Adjust phpunit.xml.dist: fixate UTC timezone for local dev usage.

Fix & update stylelint. (also solves #1203)


packages/typo3-docs-theme/assets/js/all-documentations-menu.js
```diff
@@ -133,13 +113,9 @@ class AllDocumentationsMenu extends AllDocumentationsMenuBase {
 
   createTooltip() {
     const element = document.createElement('div');
+    element.id = 'all-docs-tooltip';
+    element.setAttribute('popover', '');
     element.classList.add(this.createClassName('tooltip'));
-    element.setAttribute('role', 'tooltip');
```
ineswillenbrock, line 137: please keep this line, seems to be still needed for a11y

packages/typo3-docs-theme/assets/sass/components/directives/_admonition.scss
```diff
@@ -1,42 +1,36 @@
 @import "../../variables";
-
```
ineswillenbrock, line 2: Please let this newlines in the files: it allows for easier reading and editing.

packages/typo3-docs-theme/assets/sass/components/directives/_versionchanged.scss
```diff
@@ -25,9 +26,10 @@
 .versionchanged {
     @include version-changed($info);
 }
-
 .admonition {
-    .deprecated, .versionadded, .versionchanged {
+    .deprecated,
```
ineswillenbrock, line 37: I actually like this style change but check the style guide or editorconfig of this repository or the core

packages/typo3-docs-theme/assets/sass/components/_highlightjs_github.scss
```diff
@@ -8,31 +8,29 @@
   Outdated base version: https://github.com/primer/github-syntax-light
```
ineswillenbrock, line 8: This file is a prime example why one should not delete empty lines :)

packages/typo3-docs-theme/assets/sass/components/_rst.scss
```diff
@@ -22,8 +22,6 @@
     //    margin: 0;
     //}
 
-
-
```
ineswillenbrock, line 26: one newline should be fine here

packages/typo3-docs-theme/assets/sass/components/_search.scss
```diff
@@ -6,68 +6,58 @@
     width: auto;
     display: flex;
     justify-content: center;
-
     &::backdrop {
-        background-color: rgba(0, 0, 0, 0.5);
+        background-color: rgba(0, 0, 0, .5);
```
ineswillenbrock: Please keep the 0

packages/typo3-docs-theme/assets/sass/components/_textroles.scss
```diff
@@ -1,5 +1,6 @@
 
 article {
+
```
ineswillenbrock, line 3: Why did you add this newline but deleted the one below?

packages/typo3-docs-theme/assets/sass/mixins/_focus.scss
```diff
@@ -1,3 +1,3 @@
 @mixin focus() {
-    box-shadow: inset 0px 0px 2px 2px;
+    box-shadow: inset 0 0 2px 2px;
```
ineswillenbrock, line 2: I'm fine with dropping px

packages/typo3-docs-theme/assets/sass/_variables.scss
```diff
@@ -32,7 +32,7 @@ $secondary:         #333333;
 $secondary-dark:    shade-color($secondary, 20%);
 $secondary-light:   tint-color($secondary, 20%);
 
-$tertiary:          #005E85;
+$tertiary:          #005e85;
```
ineswillenbrock, line 35: fine with me to switch to lowercase

packages/typo3-docs-theme/.stylelintrc
```diff
@@ -1,97 +1,101 @@
 {
+  "customSyntax": "postcss-scss",
```
ineswillenbrock, line 2: why did you change this file?

ineswillenbrock (changes requested):
Please add at least the role="tooltip" to All Documentation as mentioned.
And let's discuss the styling changes in the team as those enlage this PR.

ineswillenbrock:
Note to everyone: after merging and releasing this PR we need to check api.typo3.org: popper is linked there and should get removed as well as the version in the CDN paths adopted.
The next content change there after the release might break things.

richardkrikler:
@ineswillenbrock 
> Please add at least the role="tooltip" to All Documentation as mentioned. And let's discuss the styling changes in the team as those enlage this PR.

I understand the a11y problem & pushed a commit readding the deleted line.
Not sure if role tooltip is actually the right choice here, as it is more a menu than a tooltip 🤔.

I cannot fully understand the other comments about SCSS code styling changes, as these were not done by me manually.
The existing Gruntfile.js got the wrong path for stylelint & so the config was not applied for a while.
The rules in the `.stylelintrc` were there before and are unchanged (I only migrated them to be compatible with stylelint v16 & v17 - introduction of @stylistic/stylelint-plugin was necessary here).

packages/typo3-docs-theme/.stylelintrc
richardkrikler: Please look at the commit message of 63bc15274cb34f4408be40ca10eaadbfc89e96e0
It was necessary to be compatible with v16 of stylelint.
In v16 stylistic rules were removed and reloacted into another npm package → https://github.com/stylelint-stylistic/stylelint-stylistic#about-and-purpose
https://stylelint.io/migration-guide/to-16/

ineswillenbrock (approved):
@richardkrikler regarding a11y: that's what I got reading this article: https://www.smashingmagazine.com/2026/03/getting-started-popover-api/
You mightbe right, if one refers to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tooltip_role
Might be actually rather role=menu if I read through the roles: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menu_role

Right now I'd like to keep the tooltip and open another issue to possible fix this object

ineswillenbrock (approved):
LGTM

PR #1249, opened 2026-05-12
[BUGFIX] Lookup Packagist dev versions if no stable versions were found

eliashaeussler:
When documentation is written for an unreleased Composer package on Packagist and rendering is performed with `--fail-on-log` flag, the rendering fails due to the following warning:

```
[2026-05-12T06:47:16.166358+00:00] app.WARNING: "cpsit/typo3-handlebars-forms" was not found on packagist.  {"rst-file":"Index.rst"} []
```

The message is a bit misleading, because the package actually exists, but does not contain a stable version yet.

Therefore, this PR extends Composer package lookup by an additional lookup for dev versions if no stable versions are found. The first lookup (e.g. https://repo.packagist.org/p2/cpsit/typo3-handlebars-forms.json) is extended by a second lookup (https://repo.packagist.org/p2/cpsit/typo3-handlebars-forms~dev.json).


linawolf (approved):

linawolf:
Thanks for contributing!

PR #1264, opened 2026-05-23
[BUGFIX] Rebuilding assets after mergin

linawolf:
Follow up for https://github.com/TYPO3-Documentation/render-guides/pull/1255

PR #1265, opened 2026-05-24
Auto-rebuild & auto-merge theme-asset Dependabot PRs (+ verify check)

CybotTM:
## Problem

`packages/typo3-docs-theme/` commits its **built** frontend assets (`assets/js/`, `resources/public/`). They drift out of sync when a contributor forgets to rebuild, or when Dependabot bumps a build dependency — Dependabot can't rebuild the assets (this is what made the manual follow-up #1264 necessary).

## Changes

1. **`check-built-assets.yaml`** — required check `verify`: on every PR, rebuilds the theme and fails if the committed assets don't match a fresh build. PRs not touching the theme short-circuit to success, so the check can be required without blocking unrelated PRs. Also runs on push to `main` and in the merge queue.
2. **`dependabot-rebuild-assets.yaml`** — on a Dependabot theme bump: rebuilds the assets, commits them back as a GitHub-signed (Verified) commit via a GitHub App, then enables auto-merge for patch/minor updates.
3. **`pr-auto-merge.yaml`** — excludes theme PRs from the generic Dependabot auto-merge; the workflow above handles them after the rebuild commit lands.
4. Docs pointer in `Documentation/Developer/ThemeCustomization.rst`.

## Required repository settings

- [ ] GitHub App with **Contents: write** + **Pull requests: write**, exposed as secrets `APP_ID` / `APP_PRIVATE_KEY` (same names as `reusable-backport.yml`).
- [ ] **Allow auto-merge** enabled.
- [ ] Branch protection requiring the **`verify`** check.

## Security

`dependabot-rebuild-assets.yaml` uses `pull_request_target` (same convention as `pr-auto-merge.yaml`). Guards: actor gated to `dependabot[bot]`; checkout pinned to the PR head SHA; lockfile-only diff guard; `npm ci --ignore-scripts`; read-only `GITHUB_TOKEN`; App token minted only **after** the build; commit created via `createCommitOnBranch`, so it is Verified and re-triggers `verify`.

Residual risk (accepted): patch/minor theme bumps merge without a human in the loop, so a compromised dependency's build-time code runs and its output merges. For a manual checkpoint, drop the "Enable auto-merge" step.

## Validation

Proven end-to-end on a fork: non-theme PR → `verify` skips green; stale assets → `verify` blocks; Dependabot bump → Verified rebuild commit → `verify` re-runs → auto-merge with correct assets on `main`.

### DIFF
--- a/.github/workflows/check-built-assets.yaml
+++ b/.github/workflows/check-built-assets.yaml
@@ -1,0 +1,116 @@
+# Contract workflow: fail a PR whose committed theme assets diverge from a
+# fresh build.
+#
+# This is the actual gate — mark it as a required status check in branch
+# protection. It runs on EVERY pull request (no paths filter) so that the
+# required check always reports a status: a required check that is skipped by
+# a paths filter never reports on unrelated PRs, which leaves them blocked
+# forever ("Expected — waiting for status"). PRs that don't touch the theme
+# package short-circuit to success without building.
+#
+# The companion dependabot-rebuild-assets.yaml is the convenience layer that
+# keeps Dependabot PRs from tripping this check.
+
+name: "Check built assets"
+
+on: # yamllint disable-line rule:truthy
+  pull_request:
+
+permissions: {}
+
+jobs:
+  verify:
+    name: "verify"
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: "Harden runner"
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          egress-policy: block
+          disable-sudo: true
+          allowed-endpoints: >
+            api.github.com:443
+            github.com:443
+            objects.githubusercontent.com:443
+            release-assets.githubusercontent.com:443
+            registry.npmjs.org:443
+
+      - name: "Checkout"
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+
+      # Skip the (relatively expensive) build for PRs that don't touch the
+      # theme package. The job still completes successfully, so the required
+      # "verify" status is reported on every PR.
+      - name: "Detect theme-package changes"
+        id: scope
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          if git diff --name-only "$BASE_SHA" HEAD -- packages/typo3-docs-theme/ | grep -q .; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+            echo "No changes under packages/typo3-docs-theme/ — asset check not applicable."
+          fi
+
+      - name: "Install Node"
+        if: steps.scope.outputs.changed == 'true'
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version-file: "packages/typo3-docs-theme/package.json"
+          cache: "npm"
+          cache-dependency-path: "packages/typo3-docs-theme/package-lock.json"
+
+      - name: "Install npm dependencies (no install scripts)"
+        if: steps.scope.outputs.changed == 'true'
+        working-directory: packages/typo3-docs-theme
+        run: "npm ci --ignore-scripts"
+
+      - name: "Build assets"
+        if: steps.scope.outputs.changed == 'true'
+        working-directory: packages/typo3-docs-theme
+        run: "npm run build"
+
+      - name: "Fail if generated assets differ from committed output"
+        if: steps.scope.outputs.changed == 'true'
+        run: |
+          ASSET_PATHS=(
+            packages/typo3-docs-theme/assets
+            packages/typo3-docs-theme/resources/public
+          )
+
+          # Two failure modes to detect:
+          #   1. Tracked files changed (git diff)
+          #   2. New untracked files were generated under the asset trees
+          #      (git diff alone would miss these)
+          modified="$(git diff --name-only -- "${ASSET_PATHS[@]}")"
+          untracked="$(git ls-files --others --exclude-standard -- "${ASSET_PATHS[@]}")"
+
+          if [ -z "$modified" ] && [ -z "$untracked" ]; then
+            exit 0
+          fi
+
+          # Full diagnostics go to the log; the ::error:: annotation is a
+          # single concise line so the PR Files-changed view doesn't get
+          # cluttered with one annotation per echo.
+          echo "Files out of sync with current build:"
+          if [ -n "$modified" ]; then
+            printf '%s\n' "$modified" | sed 's/^/  modified: /'
+          fi
+          if [ -n "$untracked" ]; then
+            printf '%s\n' "$untracked" | sed 's/^/  untracked: /'
+          fi
+          echo ""
+          echo "Reproduce locally:"
+          echo "  cd packages/typo3-docs-theme && npm ci && npm run build"
+          echo "  git add assets resources/public && git commit"
+          echo ""
+          echo "See Documentation/Developer/ThemeCustomization.rst for context."
+          echo "::error::Generated assets are out of date. Run 'npm ci && npm run build' in packages/typo3-docs-theme/ and commit the regenerated assets. See run log for affected files."
+          exit 1
--- a/.github/workflows/check-built-assets.yaml
+++ b/.github/workflows/check-built-assets.yaml
@@ -8,6 +8,15 @@
 # forever ("Expected — waiting for status"). PRs that don't touch the theme
 # package short-circuit to success without building.
 #
+# The push trigger is not redundant with the pull_request one: security
+# advisory fixes are merged out of a temporary private fork and arrive on
+# "main" as direct commits without ever being a pull request here, so the
+# pull_request trigger alone cannot see them. That is how the 0.40.0 assets
+# came to be published without the fixes they were tagged for.
+#
+# merge_group is required because "main" uses a merge queue: without it,
+# queued merges never receive this required status check and time out.
+#
 # The companion dependabot-rebuild-assets.yaml is the convenience layer that
 # keeps Dependabot PRs from tripping this check.
 
@@ -15,6 +24,10 @@ name: "Check built assets"
 
 on: # yamllint disable-line rule:truthy
   pull_request:
+  push:
+    branches:
+      - "main"
+  merge_group:
 
 permissions: {}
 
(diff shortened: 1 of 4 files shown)


garvinhicking:
So, "allow auto merge" is enabled. The App secrets should exist. I'm not sure what you mean with "verify status" though.

The only "verify" I can find under branch protection is the one here:

Screenshot 2026-05-24 at 20 14 50

and I don't think we can require that...

CybotTM:
@garvinhicking 

"verify" status check means the "verify" job:

https://github.com/TYPO3-Documentation/render-guides/pull/1265/changes#diff-31e507d3ed5388405d0382e860458b4c02af7444aa1084a8aed89f3c5da70d36R22

It must be required to ensure the PR does not get (auto-)merged when it fails.

image

You may only be able to set it after merge.

And regarding the "signed commits", IMO they should be required along with signed-off commits.
`git commit --gpg-sign --signoff ...` (git commit -s -S ...) is not that hard.

You can easily enable it for all commits:

```
git config --global gpg.format ssh
git config --global user.signingkey ~/.ssh/id_ed25519.pub
git config --global commit.gpgsign true
```

after registering your ssh key in GitHub → SSH and GPG keys → Signing keys
and your commit email `git config user.email` must match one of your verified emails in GitHub.

CybotTM:
I will check failing "[Main / Validate monorepo structure (pull_request)](https://github.com/TYPO3-Documentation/render-guides/actions/runs/26387644383/job/77669687208?pr=1265)Failing after 36s"

CybotTM:
Rebased onto `main` and extended.

The `[BUGFIX] Use the valid 'ignore' key in dependabot.yml` commit dropped out of the branch — #1346 has since landed the same `ignore` block on `main`, so `.github/dependabot.yml` is now identical to it and this PR no longer touches that file.

Two gaps the branch had, both from things that happened after it was opened in May:

**Advisory merges bypassed it.** Fixes from a temporary private fork arrive on `main` as direct commits, never as a pull request here, so a `pull_request`-only trigger cannot see them. That is not hypothetical — it is how 0.40.0 came to ship theme assets built eleven days before the fixes it was tagged for (#1353).

**The merge queue would have stalled.** `main` gained a merge queue in #1326. A required status check without a `merge_group` trigger never reports on a queued merge, so every queued merge times out — the same failure mode this file's own header warns about for paths filters.

Both triggers are added. The base sha now comes out of the event payload via `jq` instead of being interpolated into the script, and an event that carries no base sha — a push — is verified in full rather than skipped.

Checked locally: `actionlint` clean, `actionlint -shellcheck` clean, `yamllint` reports the same 11 findings as before the change. The base-sha extraction was exercised against synthetic `pull_request`, `merge_group` and `push` payloads, which resolve to the PR base, the merge-group base, and full verification respectively.

linawolf:
Thanks for this — I went through it properly, including reproducing the build
locally to test the premise. The design holds up: `npm ci --ignore-scripts &&
npm run build` on the committed lockfile gives me **zero** diff and **zero**
untracked files under both asset trees, so `verify` is enforceable as it
stands, not aspirational. The reasoning in the comments is also genuinely
useful to read — the "a required check must never be paths-skipped" point and
the `push`-trigger rationale for advisory merges are both correct and not
obvious.

Four things I'd like to resolve before this goes in.

## 1. The `pr-auto-merge.yaml` hunk predates the merge queue

This is the main one, and it's a timing artefact rather than a mistake:

- 2026-05-24 — this PR opened, including the `pr-auto-merge.yaml` change (5d84182)
- 2026-07-18 — #1326 added the `merge_group` trigger; `main` went behind a merge queue
- 2026-07-25 — you rebased and covered the queue **in the asset check** (567560a, c27d2dc)

Both July commits touch only `check-built-assets.yaml` and
`dependabot-rebuild-assets.yaml`. `pr-auto-merge.yaml` is still the May version,
and its rationale —

> if auto-merge fires here at PR-open time it squashes BEFORE the rebuild
> commit lands

— described the repo accurately in May, when `--squash --auto` really did merge
directly. Under a merge queue it no longer does: enabling auto-merge *enqueues*,
and the merge group re-runs the required checks, so the rebuild commit would
already be on the branch by the time anything merges.

Which raises the question I'd most like your read on: **is the theme exclusion
still needed at all?** If the queue won't merge until `verify` passes on the
rebuilt commit, the generic path may handle theme PRs correctly on its own, and
this hunk plus the `::notice::` step could just be deleted. You designed it —
tell me if there's a race I'm not seeing.

Related: #1399 rewires the same file, because enqueuing with `GITHUB_TOKEN`
produces a merge group that `Main` never runs on (token-created events don't
trigger workflows), so entries sit until `check_response_timeout` drops them.
That's visible in the queue right now — five Dependabot PRs, approved, `CLEAN`,
"Dependabot auto-merge" green, unmerged for two weeks. Worth noting your
rebuild workflow **already gets this right** and enqueues with the App token;
it's only the generic path that still uses `GITHUB_TOKEN`.

Suggested order: land #1399 first, then rebase this onto it, revisiting the
exclusion rather than resolving the conflict mechanically.

Same era, same file family: `gh pr merge --squash --auto` in
`dependabot-rebuild-assets.yaml` — under a merge queue the method comes from the
queue rule and `gh` warns the flag is ignored. Dropping `--squash` matches what
#1399 does.

## 2. Minting the token after the build doesn't close the hole

The security note argues the App token is safe because it's minted after
`npm run build`, so untrusted build code can't read it. That closes the direct
read, but not the job.

A build step can append to `$GITHUB_ENV` or `$GITHUB_PATH` — ordinary writable
files whose contents GitHub applies to *subsequent* steps. `NODE_OPTIONS=--require
/tmp/x.js` in `$GITHUB_ENV` is picked up by `actions/create-github-app-token` and
`actions/github-script`, both Node actions, both of which then hold a
Contents:write + PRs:write token. The final `gh pr merge` step exports `GH_TOKEN`
into a shell that a hijacked `PATH` can intercept.

The fix realises what the current ordering is already reaching for: split into
two jobs. Job A checks out, builds, uploads the asset trees as an artifact —
no secrets in scope at all. Job B downloads the artifact, mints the token,
commits. Untrusted code and the token then never share a runner.

I'd rather do that than accept it as residual risk. "Patch/minor theme bumps
merge without a human in the loop" is exactly the case where the runner should
be assumed hostile.

## 3. Node version: the new workflows disagree with the existing one

```
main.yaml (JS tests)            node-version-file: .nvmrc        -> 24.16.0 exactly
check-built-assets.yaml         node-version-file: package.json  -> engines ">=24.16.0 <25.0.0"
dependabot-rebuild-assets.yaml  node-version-file: package.json  -> same range
```

setup-node resolves the range to the newest 24.x — 24.18.0 today. For a check
whose whole job is byte-comparing build output, the toolchain should be pinned
to what a contributor gets from `.nvmrc`. Not breaking right now (I built on
24.18.0 and got a clean diff), but it's a latent flake source aimed straight at
this PR's purpose. Two-line fix.

## Smaller notes

- **Three texts, three stories.** The new `::notice::` says theme PRs "require a
  manual review/merge"; `dependabot-rebuild-assets.yaml` enables auto-merge for
  patch/minor; the docs say "no manual rebuild is required there." Worth
  reconciling, especially if item 1 removes the exclusion.
- **`verify` is a very generic required-check name**, and `main.yaml` carries an
  explicit warning that job names are wired into the branch ruleset.
  `check-built-assets` would be safer to live with.
- **The untracked-file hard-fail may be fragile under the bumps it automates.**
  `vite.config.js` uses `chunkFileNames: 'js/[name].[hash].js'` with
  `emptyOutDir: false`. Today the search bundle is one chunk, so nothing hashed
  is emitted. If a vite or react bump ever triggers code-splitting, the new
  hashed file lands untracked, "Prepare commit inputs" refuses to commit, and
  superseded hashed files would never be cleaned up. Speculative, but the
  mechanism is concrete and vite/react bumps are precisely the traffic here.
- `dependabot/fetch-metadata` runs with only `contents: read` — worth confirming
  it doesn't want `pull-requests: read`. It would fail loudly, so low risk.

## What I verified locally

- `npm ci --ignore-scripts` + `npm run build` on node 24 -> clean `git status`;
  no modified or untracked files under `assets/` or `resources/public/`.
- `--ignore-scripts` doesn't break the build (rolldown and sass need no install
  scripts).
- `steps.metadata.outputs.directory` is `/packages/typo3-docs-theme/` for the npm
  ecosystem, so the `contains(...)` exclusion matches; composer and
  github-actions both use `/` and are unaffected.
- `main.yaml` is the only workflow with a `merge_group` trigger today, so adding
  one to `check-built-assets.yaml` is right.

The asset-drift problem is real and this is the right shape of solution — I'd
like to get it in. Item 1 is the one I need your answer on; 2 and 3 I'm happy to
push to your branch if that's easier than round-tripping.

CybotTM:
Thank you for reproducing the build — that is the part I could not ask for. All four points are addressed; item 1 is the one where I have a measurement that changes the answer.

## 1. The theme exclusion is still needed, for a different reason than the comment gave

Your reading of the merge queue is right, and the May rationale in that comment was indeed out of date. But the conclusion does not follow yet, because the queue only holds a pull request for the checks the ruleset *requires*, and the asset check is not one of them. `merge-queue-main` requires seven: `Tests (8.2)`, `Tests (8.3)`, `Tests (8.4)`, `Tests (8.5)`, `Quality`, `typo3-docs-theme JS tests`, `Validate monorepo structure`.

So today the sequence is: Dependabot opens a theme PR, the seven required checks go green on the stale assets, `pr-auto-merge` enqueues it, and the queue merges it — the asset check is reporting, but nothing waits for it. The exclusion is what prevents that, not the squash-versus-enqueue distinction.

It becomes deletable the moment `check-built-assets` is added to the required set, and the comment now says exactly that, so whoever adds it finds the instruction to remove the exclusion and the notice. If you would rather do it the other way round — add the check to the ruleset first and drop the exclusion in this PR — say so and I will; it needs an admin on the ruleset either way, and I did not want to make that call inside a PR that can only change files.

Two related things from your note are done: `--squash` is gone from the rebuild workflow's `gh pr merge` (the queue rule sets the method; it is `SQUASH`), and the merge-queue-aware path from #1399 is what this branch now carries after the rebase.

## 2. Two jobs, so the token and the build never share a runner

You are right that minting after the build closes only the direct read. `$GITHUB_ENV` and `$GITHUB_PATH` are the hole, and both reach Node actions.

`build` checks out, builds, and uploads the two asset trees plus the path lists as an artifact — no secret in scope. `commit` downloads the artifact, mints the token and writes the commit through the API; it has no checkout step and no npm, so nothing from the dependency tree executes beside the token. The artifact crosses as data. `expectedHeadOid` still comes from the build job, so a force-push between the two jobs fails the mutation rather than being overwritten by it.

## 3. Node pin

Both workflows read `packages/typo3-docs-theme/.nvmrc` now, the same file `main.yaml` uses, instead of the engines range.

## 4. Smaller notes

- **Three texts** now say one thing: the rebuild workflow commits the assets and enables auto-merge itself for a patch or minor bump, and a failed build leaves the PR open. The `::notice::`, the docs paragraph and the exclusion comment were each rewritten to that.
- **`verify`** is now `check-built-assets`, job id and name — free to rename while it is not yet required.
- **Untracked files** are committed instead of refused. Your code-splitting scenario is the reason: the first hashed chunk would arrive untracked and the old step would have stranded the PR. A change *outside* the two asset trees still fails the job, which is what that guard was for. Stale hashed files from an earlier build are a separate matter — they would need `emptyOutDir` or a clean step, and I would rather not change the vite config in this PR.
- **`fetch-metadata` permissions:** the new `commit` job declares `pull-requests: read`. I left `pr-auto-merge.yaml`'s permissions alone, since that file's job is main's and works today.

What I could not verify: the workflow itself. It needs a Dependabot PR and the App secrets, so there is no local run behind it — only `actionlint` (clean on all three files) and a replay of the new guard's pathspec against a dirty tree, where a file under the asset trees is not reported and one outside them is. The first real Dependabot theme bump after this merges is the actual test, and the asset check catches it if the rebuild misbehaves.

Happy to take items in another shape if you prefer — and if you would rather push the ruleset change and have me delete the exclusion, that is one commit.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01RxZ2DkBeChDfVViLxvazbZ)_

CybotTM:
Correction to my previous comment, item 2: the `NODE_OPTIONS` example I gave does not work, and I had not checked it.

The runner keeps a block list for the `$GITHUB_ENV` file command, and `NODE_OPTIONS` is its only entry — `_setEnvBlockList` in `src/Runner.Worker/FileCommandManager.cs`, compared case-insensitively, the pair skipped with a message. So that specific variable never reaches `create-github-app-token` or `github-script`. A CodeRabbit review on the skill documentation where I wrote the same thing up flagged it, and I read the runner source rather than take either side on trust.

What the argument rests on is unchanged, because `$GITHUB_PATH` is not blocked: a directory prepended there intercepts the `gh` that the last step calls with `GH_TOKEN` exported, and the same goes for any other variable a later step or action reads. The two-job split is still the fix, and the blocked route is worth naming precisely because it is the first example anyone reaches for.

Also corrected in the same commit ([`5e2ea171`](https://github.com/TYPO3-Documentation/render-guides/pull/1265/commits/5e2ea171)): the header comment said the build job "holds no secret at all". `contents: read` still issues a `GITHUB_TOKEN`, and `actions/checkout` would write it into `.git/config` — what keeps it off that runner is `persist-credentials: false`, which the job already sets. The comment now says that instead of claiming there is nothing to steal.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01RxZ2DkBeChDfVViLxvazbZ)_

linawolf (approved):
Looks Good to me, thank you

PR #1291, opened 2026-06-16
[BUGFIX] Keep navigation menu in toctree order with external links

### LINKED ISSUE #1175: Different sorting in menu than in toctree
I have a `Index.rst` with the following table of contents:

```rst
..  toctree::
    :titlesonly:

    Introduction
    Fluid Syntax <https://docs.typo3.org/permalink/fluid:syntax>
    UsingFluidInTypo3
    UsingTypoScriptCObjectViewHelper
    AdditionalAttributes
    DevelopCustomViewhelper
    ViewHelper reference <https://docs.typo3.org/other/typo3/view-helper-reference/main/en-us/>
```

On the page itself, everything is in the right order. However, in the sidebar navigation, the external links are reordered to the top of the list. See here:

https://docs.typo3.org/m/typo3/reference-coreapi/13.4/en-us/ApiOverview/Fluid/Index.html

CybotTM:
Fixes #1175

## Problem

When a `toctree` mixes internal pages and external links (e.g. interlinks to other manuals), the **sidebar navigation** renders all external entries grouped at the top, even though the on-page `toctree` keeps the authored order. On the reported page (`reference-coreapi` → Fluid) the sidebar shows `Fluid Syntax` and `ViewHelper reference` (both external) above the internal pages.

## Root cause

The guides compiler attaches a document's menu entries via **separate transformers for internal and external entries** (`InternalMenuEntryNodeTransformer` / `ExternalMenuEntryNodeTransformer`), and the compiler runs **one full tree traversal per transformer** at each priority. So every external entry is attached to the document's menu-entry list in one pass and every internal entry in another — the list ends up grouped by type instead of following the authored toctree. The on-page toctree renders from the `TocNode` (authored order) and stays correct; the sidebar renders from `DocumentEntryNode::getMenuEntries()` and does not.

This is an engine-level issue in `phpdocumentor/guides`. The proper fix belongs upstream (phpDocumentor/guides#1344); this PR adds a **downstream compiler pass** so docs.typo3.org is fixed now, and it can be removed after a future `guides` bump.

## Fix

`SortMenuEntriesByToctreeTransformer` realigns each document's menu entries with the authored order of its `toctree`s, matching entries by file/URL. It emits each toctree's entries as one contiguous block at the position of its first entry, so a document with **several** mixed toctrees is ordered correctly too. **Globbed toctrees are skipped** (their order comes from the glob expansion, which the sitemap relies on).

## Before / after

Sidebar of a subpage whose toctree mixes internal pages with two external links:

**Before**

before

**After**

after

## Verification

- Two integration fixtures: `tests-full/menu-external-order` (one nested mixed toctree) and `tests-full/menu-multiple-toctrees-order` (two mixed toctrees on one page). Each **passes with the fix and fails without it**.
- Full integration suite green (116 tests); the change touches only mixed/explicit toctrees — `external-menu`, `two-toctrees`, `menu-subpages`, `sitemap` (glob) and `main-menu-json` are unaffected.
- PHPStan (repo config) and php-cs-fixer clean on the changed files.


CybotTM:
looking into failing CI ...

linawolf:
@CybotTM tests are still failing

CybotTM:
> @CybotTM tests are still failing

hey, not fair, they are failing **again**, not still ... ;-)

CybotTM:
main fixed and PR rebased.

linawolf (approved):

PR #1292, opened 2026-06-16
[BUGFIX] Sort version switcher correctly for versions like 0.10

CybotTM:
## Problem

For a package with both `0.1` and `0.10` (e.g. [netresearch/nr-vault](https://docs.typo3.org/p/netresearch/nr-vault/0.10/en-us/)) the version switcher is wrong in two ways:

- **`0.10` is sorted last** instead of first — it should appear right after `main`.
- **Opening the `0.10` page pre-selects `0.1`** in the dropdown.

## Cause

`versions.js` sorted by `parseFloat(v)`, and `parseFloat("0.10") === 0.1`, so `0.10` ties with the `0.x` group and lands at the bottom. Pre-selection compared against the rendered `data-current-version` attribute, which is `"0.1"` on the `0.10` page because the version string `"0.10"` is numerically coerced to `"0.1"` server-side (a separate, deeper bug in the render pipeline).

## Fix

- Sort each dotted version component **numerically** (`main` first, then highest version), so `0.10` > `0.9` > … > `0.1`.
- Derive the active version from the **page URL** (the authoritative source, e.g. `…/0.10/en-us/…`) instead of the coercible `data-current-version` attribute. This makes pre-selection correct regardless of that server-side coercion.

`resources/public/js/theme.min.js` is rebuilt (grunt uglify).

## Before / after

The version dropdown on the `nr-vault` `0.10` page:

**Before**

before

**After**

after

## Tests

New `tests/js/versions.test.js` (vitest/jsdom) asserts both the sort order (`main, 0.10, 0.9, … 0.1`) and that the `0.10` page pre-selects `0.10`. Both **pass with the fix and fail without it**.

## Related

The wrong pre-selection has a second, deeper cause that this PR does **not** rely on: the `0.10` page is served with `data-current-version="0.1"` (and title / meta `0.1`) because `guides.xml`'s `version="0.10"` is numerically coerced to the float `0.1` while parsing (Symfony `XmlUtils` phpize). That is tracked in #1293 and fixed upstream in phpDocumentor/guides#1345. This PR makes the switcher correct regardless — it derives the active version from the page URL — so the two fixes are independent.


packages/typo3-docs-theme/tests/js/versions.test.js
```diff
@@ -0,0 +1,98 @@
...
+      'https://docs.typo3.org/p/netresearch/nr-llm/0.12/en-us/Index.html', '0.12', NR_LLM_VERSIONS);
+    const options = await runAndWait(count);
+
+    expect(options.map(option => option.textContent)).toEqual(
+      ['main', '0.12', '0.11', '0.10', '0.9', '0.8', '0.7', '0.6', '0.5', '0.4', '0.3', '0.2', '0.1']);
+    expect(options.find(option => option.selected).textContent).toBe('0.12');
+  });
+
+  it('orders versions with a differing number of components (13.4.21 > 13.4 > 13.3)', async () => {
+    const versions = versionList(['13.3', '13.4', '13.4.21', '14.0', 'main']);
```
linawolf, line 120: It might make sense to have a test with version 9.5 and 8.7 in ti to demonstrate their order is resolved

CybotTM: Done

linawolf (approved):

PR #1294, opened 2026-06-17
[BUGFIX] Strip leading backslash from :php-short: output

### LINKED ISSUE #1155: Incorrect output from :php-short:
I've noticed that some PHP class names are outputted incorrectly if the :php-short: syntax is used. I don't think that any of the examples listed here should have a leading slash, as this would refer to a different PHP class:
https://docs.typo3.org/permalink/h2document:links-api-short

CybotTM:
## Problem

The `:php-short:` text role renders class names with a leading backslash for classes that are **not** resolved via the TYPO3 API — e.g. `\LoggerInterface` instead of `LoggerInterface`, `\Something` instead of `Something`. As noted in #1155, a leading backslash changes the meaning (it would refer to a root-namespace class), so the output is incorrect.

rg1294-before-fix

## Cause

In `PhpTextRole::processNode()` the short name was taken from `$fqn[2]`, the last match of the repeated capture group in `CLASS_NAME_PATTERN_REGEX`. That group matches `\Segment`, so the captured value keeps the leading namespace separator. When the class is found in the API, the `apiInfo['short']` value overrides it — which is why the bug only surfaced in the fallback branches (PSR, Fluid, MyVendor, generic).

## Fix

Strip the leading backslash from the short name (falling back to the full name when there is no inner segment). One-line change in the role plus a regression test.

rg1294-after-fix

## Test

Adds integration test case `tests/Integration/tests/code/code-php-short` covering the fallback rendering. Without the fix the rendered output is `\FooBar` / `\Foo<wbr>Bar`; with it, `FooBar` / `Foo<wbr>Bar`.

Verified locally: integration + unit suites pass, PHPStan clean, php-cs-fixer clean.

Fixes #1155

### DIFF
--- a/packages/typo3-docs-theme/src/TextRoles/PhpTextRole.php
+++ b/packages/typo3-docs-theme/src/TextRoles/PhpTextRole.php
@@ -46,7 +46,8 @@ final class PhpTextRole implements TextRole
             $apiInfo = $this->typo3ApiService->getClassInfo($rawContent);
             $name = $rawContent;
             if ($role === 'php-short') {
-                $name = $fqn[2] ?? $rawContent;
+                $shortName = $fqn[2] ?? '';
+                $name = ltrim($shortName !== '' ? $shortName : $rawContent, '\\');
             }
             return $this->getClassCodeNode($rawContent, $apiInfo, $role, $name, $type);
         }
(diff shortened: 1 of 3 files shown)


garvinhicking (approved):

garvinhicking:
I think this looks right, but not feeling safe enough to merge since I never used the directive enough on my own yet :D

ineswillenbrock (approved):
LGTM.

Rendered https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument with a custom container and the result fullfills the expectation of Issue https://github.com/TYPO3-Documentation/render-guides/issues/1155

PR #1297, opened 2026-06-17
[BUGFIX] Fix FontAwesome icons breaking in Firefox with scroll-timeline polyfill

benjaminkott:
The scroll-timeline polyfill re-serves same-origin stylesheets from a blob: URL, which has no path context, so theme.css's relative webfont url()s no longer resolve. The icon font fails to load and icons render as empty placeholder boxes in Firefox.

Patch the polyfill at build time to absolutize relative url()s against the stylesheet's original href before it creates the blob, and drop the crossorigin="anonymous" attribute that broke CSS loading from the CDN (cross-origin sheets need ACAO when fetched in CORS mode). Also strip the polyfill's dangling sourcemap comment.


benjaminkott:
abandoned in favor of https://github.com/TYPO3-Documentation/render-guides/pull/1298

PR #1301, opened 2026-06-20
[TASK] Cleanup Gruntfile.js

richardkrikler:
* Remove unused path const
* Simplify removesourcemap task
* Remove unused config entry pkg
* Remove non-functional copy:css task block (theme.css gets placed into public/css via sass:build task block)
* Remove non-existant files from sass:build/debug task block
  9a5964782c94376b4ad05e17c1dea36f893a9d49 "[TASK] Integrate all css into one file"
* Remove "update"-task declaration & directly use "copy" instead

### DIFF
--- a/packages/typo3-docs-theme/Gruntfile.js
+++ b/packages/typo3-docs-theme/Gruntfile.js
@@ -1,24 +1,20 @@
 module.exports = function (grunt) {
 
-  const path = require('path');
   const sass = require('sass');
 
   /**
    * Grunt task to remove source map comment
    */
   grunt.registerMultiTask('removesourcemap', 'Grunt task to remove sourcemp comment from files', function () {
-    var done = this.async(),
-      files = this.filesSrc.filter(function (file) {
-        return grunt.file.isFile(file);
-      }),
-      counter = 0;
+    const done = this.async();
+    let counter = 0;
     this.files.forEach(function (file) {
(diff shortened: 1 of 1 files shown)


CybotTM (approved):
Verified as a build-output no-op:

- Controlled A/B: full `npm run build` (vite + grunt) with the pre-PR and the PR Gruntfile on the same tree — `resources/public/` and the `copy:debug` output (`Documentation-GENERATED-temp/_resources`) are recursively byte-identical, the working tree stays clean (committed artifacts regenerate exactly), both builds exit 0.
- `path` const and `pkg` config entry: no other references in the file.
- `copy:css`: `cwd: 'css'` does not exist in the package; the build log confirms the task copied 0 files.
- Removed `sass` entries: `codeblock.scss`/`fontawesome.scss`/`webfonts.scss` were folded into `theme.scss` in 9a5964782c94376b4ad05e17c1dea36f893a9d49 and grunt-sass has been silently skipping the missing sources since; `resources/public/css/` contains only `theme.css`, so no orphaned artifacts reference the removed targets.
- `removesourcemap` rewrite: equivalent for the actual config (single target `contrib`, one 1:1 mapping — `bootstrap.min.js` onto itself).
- `update` was a pure alias for `copy`; bare `copy` runs the same targets either way.

CybotTM:
looks good to me

PR #1302, opened 2026-06-20
[TASK] Remove unused typo3-docs-theme dev-dependency

richardkrikler:
Remove unused grunt-exec npm package.


CybotTM (approved):
looks good to me -`grunt-exec` is and was never used

PR #1303, opened 2026-06-20
[FEATURE] Add :changelog: option to version change directives

CybotTM:
## What

Adds a `:changelog:` option to the `versionadded`, `versionchanged` and `deprecated` directives that renders a standardized "See changelog entry" link in the version badge. Without the option the blocks render exactly as before.

The option value supports three forms:

```rst
.. TYPO3 core change -> bare core changelog entry id
..  versionchanged:: 14.0
    :changelog: feature-107628-1729026000

.. extension change -> the extension's interlink shortcode + anchor
..  versionchanged:: 2.0
    :changelog: acme/acme-blog:changes-2-0-0

.. the current manual's own changelog -> short local form
..  versionchanged:: 2.1
    :changelog: #changes-2-1-0
```

This makes the standardized changelog link available to **extension documentation**, not only the TYPO3 core docs.

## Why

Authors regularly want to point a version-change note at the changelog entry that explains the change. Today the only way is to hand-write a link in the body, and a common mistake is to try to combine it with a role (e.g. `` :guilabel:`Administration <...>`_ ``), which does not produce a link and leaks the raw URL into the badge (see TYPO3-Documentation/TYPO3CMS-Tutorial-GettingStarted#824 for a real example). A dedicated option makes the intent explicit and avoids the broken-role mistake, and makes the standardized link available to extension docs.

## How

The change is additive and TYPO3-specific:

- `AbstractTypo3VersionChangeDirective` (+ `Typo3VersionAddedDirective`, `Typo3VersionChangedDirective`, `Typo3DeprecatedDirective`) override the phpDocumentor directives by name (same mechanism as the existing `FigureDirective` override). They read the `:changelog:` option and build a `Typo3VersionChangeNode`.
- The option value becomes an interlink reference that is **resolved against the changelog inventory during rendering**: a bare id resolves against the core `changelog` inventory; `vendor/package:anchor` against that extension's inventory; `#anchor` as a local reference to this manual's own labels. A resolvable entry renders the real target URL; a **non-existent entry logs a build warning** and is marked unresolved by the theme's own `invalid-link` markup, the same as any other unresolvable cross-reference — so a dead permalink can no longer be produced.
- The `<shortcode>:<anchor>` split is done by `ExtendedInterlinkParser`, the same parser `:ref:`, `:doc:` and `ApiClassTextRole` are given, so the option cannot drift from the rest of the project's cross-reference syntax. A value the parser will not accept as `<domain>:` is reported as malformed rather than silently taken for a local label.
- The option is read as an option rather than as a string, and untrimmed. Written without a value it is parsed as the boolean flag `true`, which stringifies to `1` and would otherwise be taken for a changelog entry id; written with its value on the following line, that `1` is the prefix the parser appends the continuation to; written with a trailing space and the value on the following line, the prefix is a space instead. The flag is rejected by reading the value rather than its string form, the other two by the rule that every valid form is a single token — which is why the value is examined before it is trimmed, since trimming erases the difference between a continued value and an inline one.

The value's encoding is checked before the whitespace pattern, because `preg_match` answers `false` rather than `0` on a malformed subject: read as "no whitespace", such a value reaches the anchor normalizer, where `UnicodeString` throws — ending the whole run with no pages written and an error naming the Twig template rather than the file at fault. Invalid UTF-8 is that constructor's only throw condition, so excluding it there is what makes the rest of the method total.

One branch is deliberately without a fixture: the trailing-space form needs an `.rst` line ending in a space, and this repository's `.editorconfig` strips trailing whitespace from `.rst`, so the case would be silently rewritten by any editor honouring it. It is verified by a rendered probe and noted in the fixture; it belongs in the unit tests for this method.
- `Typo3VersionChangeNode` carries the optional reference; `body/version-change.html.twig` renders it with a single `renderNode` call, and `inline/link.html.twig` produces either the `<a>` carrying the reference node's class, or the class-less `invalid-link` marker when it cannot be resolved.

## Rendered help page

`Documentation-rendertest/Directives/versionadded.rst` — each form is shown as code, with the core form's resolved link rendered below it:

Rendered :changelog: help page

## Tests

Two integration fixtures.

`tests/Integration/tests/version-change-changelog` covers the resolved core link, a non-existent core entry (warning, unresolved marker), an external extension shortcode, the `#anchor` local form and the explicit-own-shortcode form (both resolving locally), a `#anchor` pointing at a label nothing declares, and a no-option block for backward compatibility. Alongside them the nine inputs that produce a warning and no reference at all: nothing before the colon, a shortcode carrying a character an interlink domain cannot have, a shortcode with nothing behind the colon, a bare `#`, a space inside the value, an ideographic space inside the value, an invalid UTF-8 byte, the option written without a value, and the option whose value stands on the following line. Each warning names the offending value, so the expected warning log pins every one of them to the branch that produced it.

`tests/Integration/tests/version-change-changelog-no-shortcode` is a second manual, because the two cases in it need a `guides.xml` that sets no `interlink-shortcode`: the `#anchor` form resolving without that setting, and a version-change block with no `:changelog:` at all. It asserts that no warning log is written, which is what makes the second case a guard — and it resolves entirely offline, so that assertion cannot be tripped by a failed inventory fetch.

Every guard was seen failing with the code it guards removed, and each fails only where it should. The suite was also swept with mutations covering every branch and every warning path; the cases above are what closed the ones that came back green.

`Documentation-rendertest/Directives/versionadded.rst` carries the `:changelog:` showcase.

Local run of every gate the CI job executes — `test-unit` (83), `test-integration` (119), `phpstan`, `test-xml`, `test-docs`, `test-monorepo`, `test-rendertest`, `composer normalize` — all pass. `code-style` reports one file, `node_modules/flatted/php/flatted.php`; it is gitignored and only present because npm was run locally, and the quality job installs no `node_modules`. Worth excluding in `.php-cs-fixer.dist.php`, separately from this branch.

## Follow-up

The author-facing directive reference lives in `TYPO3CMS-Guide-HowToDocument`; I'm happy to open a docs PR there describing the new option once this is accepted.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_012Zm6UrYoeXRnucAdTdLRuN)_

### DIFF
--- a/packages/typo3-docs-theme/src/Directives/AbstractTypo3VersionChangeDirective.php
+++ b/packages/typo3-docs-theme/src/Directives/AbstractTypo3VersionChangeDirective.php
@@ -11,17 +11,26 @@ use phpDocumentor\Guides\RestructuredText\Parser\BlockContext;
 use phpDocumentor\Guides\RestructuredText\Parser\Directive;
 use phpDocumentor\Guides\RestructuredText\Parser\Productions\Rule;
 use T3Docs\Typo3DocsTheme\Nodes\Typo3VersionChangeNode;
+use T3Docs\Typo3DocsTheme\Settings\Typo3DocsThemeSettings;
 
 use function array_values;
+use function explode;
 use function str_contains;
+use function str_replace;
+use function str_starts_with;
+use function substr;
 use function trim;
 
 /**
  * TYPO3 specific version of phpDocumentor's version change directives
  * (versionadded, versionchanged, deprecated).
  *
- * In addition to the default behaviour it supports a ":changelog:" option that
- * renders a link to the related TYPO3 changelog entry, e.g.:
+ * In addition to the default behaviour they support a ":changelog:" option that
+ * renders a link to the related changelog entry via the docs.typo3.org
+ * permalink service.
+ *
+ * For a TYPO3 core change, pass the changelog entry identifier (resolved against
+ * the "changelog" inventory):
  *
  * ..  versionchanged:: 14.0
  *     :changelog: feature-107628-1729026000
@@ -29,10 +38,24 @@ use function trim;
  *     This module has been moved from :guilabel:`System` to
  *     :guilabel:`Administration`.
  *
- * The option value is a changelog entry identifier (resolved against the
- * "changelog" interlink inventory via the docs.typo3.org permalink service). A
- * fully qualified interlink target such as "changelog:feature-..." is accepted
- * as well.
+ * For an extension change, pass the extension's permalink shortcode followed by
+ * the changelog entry anchor. The shortcode is normalized to the dash form the
+ * permalink service expects ("vendor/package" -> "vendor-package"), just like
+ * the "copy permalink" button on docs.typo3.org:
+ *
+ * ..  versionchanged:: 2.0
+ *     :changelog: acme/acme-blog:changes-2-0-0
+ *
+ *     The teaser field was renamed; see the changelog entry for the migration.
+ *
+ * When linking the changelog of the current manual itself, use the short
+ * "#anchor" form. The manual's own "interlink-shortcode" (from guides.xml) is
+ * used automatically, so it does not have to be repeated:
+ *
+ * ..  versionchanged:: 2.0
+ *     :changelog: #changes-2-0-0
+ *
+ *     ...
  */
 abstract class AbstractTypo3VersionChangeDirective extends SubDirective
 {
@@ -44,6 +67,7 @@ abstract class AbstractTypo3VersionChangeDirective extends SubDirective
         Rule $startingRule,
         private readonly string $type,
         private readonly string $label,
+        private readonly Typo3DocsThemeSettings $themeSettings,
     ) {
         parent::__construct($startingRule);
     }
@@ -74,9 +98,25 @@ abstract class AbstractTypo3VersionChangeDirective extends SubDirective
             return null;
         }
 
-        $target = str_contains($changelog, ':')
-            ? $changelog
-            : self::CHANGELOG_INVENTORY . ':' . $changelog;
+        if (str_starts_with($changelog, '#')) {
+            // "#anchor": the changelog of the current manual itself. Use its own
+            // interlink-shortcode (from guides.xml) so it need not be repeated.
+            $shortcode = $this->themeSettings->getSettings('interlink_shortcode');
+            if ($shortcode === '') {
+                return null;
+            }
+
+            $target = str_replace('/', '-', $shortcode) . ':' . substr($changelog, 1);
+        } elseif (str_contains($changelog, ':')) {
+            // Explicit "<shortcode>:<anchor>". Normalize the shortcode to the
+            // dash form the permalink service expects (vendor/package ->
+            // vendor-package), exactly like the "copy permalink" button.
+            $parts = explode(':', $changelog, 2);
+            $target = str_replace('/', '-', $parts[0]) . ':' . ($parts[1] ?? '');
+        } else {
+            // Bare value: a TYPO3 core changelog entry identifier.
+            $target = self::CHANGELOG_INVENTORY . ':' . $changelog;
+        }
 
         return self::PERMALINK_BASE . $target;
     }
--- a/packages/typo3-docs-theme/src/Directives/AbstractTypo3VersionChangeDirective.php
+++ b/packages/typo3-docs-theme/src/Directives/AbstractTypo3VersionChangeDirective.php
@@ -10,6 +10,7 @@ use phpDocumentor\Guides\RestructuredText\Directives\SubDirective;
 use phpDocumentor\Guides\RestructuredText\Parser\BlockContext;
 use phpDocumentor\Guides\RestructuredText\Parser\Directive;
 use phpDocumentor\Guides\RestructuredText\Parser\Productions\Rule;
+use Psr\Log\LoggerInterface;
 use T3Docs\Typo3DocsTheme\Nodes\Typo3VersionChangeNode;
 use T3Docs\Typo3DocsTheme\Settings\Typo3DocsThemeSettings;
 
@@ -25,12 +26,13 @@ use function trim;
  * TYPO3 specific version of phpDocumentor's version change directives
  * (versionadded, versionchanged, deprecated).
  *
- * In addition to the default behaviour they support a ":changelog:" option that
- * renders a link to the related changelog entry via the docs.typo3.org
- * permalink service.
+ * All three directives support a ":changelog:" option that renders a link to
+ * the related changelog entry, built as a docs.typo3.org permalink. The value
+ * is not validated against an inventory, so a typo produces a link that 404s
+ * when clicked.
  *
- * For a TYPO3 core change, pass the changelog entry identifier (resolved against
- * the "changelog" inventory):
+ * For a TYPO3 core change, pass the changelog entry identifier; it becomes a
+ * "https://docs.typo3.org/permalink/changelog:<id>" link:
  *
  * ..  versionchanged:: 14.0
  *     :changelog: feature-107628-1729026000
@@ -50,12 +52,17 @@ use function trim;
  *
  * When linking the changelog of the current manual itself, use the short
  * "#anchor" form. The manual's own "interlink-shortcode" (from guides.xml) is
- * used automatically, so it does not have to be repeated:
+ * used automatically, so it does not have to be repeated (no link is rendered,
+ * and a warning is logged, if that setting is missing):
  *
  * ..  versionchanged:: 2.0
  *     :changelog: #changes-2-0-0
  *
- *     ...
+ *     A configuration option was renamed; see the changelog for the migration.
+ *
+ * Note: the shortcode normalization cannot round-trip a vendor or package name
+ * that itself contains a hyphen (a permalink-service limitation, see
+ * https://github.com/TYPO3GmbH/site-intercept/issues/283).
  */
 abstract class AbstractTypo3VersionChangeDirective extends SubDirective
 {
@@ -68,6 +75,7 @@ abstract class AbstractTypo3VersionChangeDirective extends SubDirective
         private readonly string $type,
         private readonly string $label,
         private readonly Typo3DocsThemeSettings $themeSettings,
+        private readonly LoggerInterface $logger,
     ) {
         parent::__construct($startingRule);
     }
@@ -87,11 +95,11 @@ abstract class AbstractTypo3VersionChangeDirective extends SubDirective
             $this->label,
             $directive->getData(),
             array_values($collectionNode->getChildren()),
-            $this->buildChangelogUrl($directive->getOptionString('changelog')),
+            $this->buildChangelogUrl($directive->getOptionString('changelog'), $blockContext),
         );
     }
 
-    private function buildChangelogUrl(string $changelog): string|null
+    private function buildChangelogUrl(string $changelog, BlockContext $blockContext): string|null
     {
         $changelog = trim($changelog);
         if ($changelog === '') {
@@ -103,21 +111,38 @@ abstract class AbstractTypo3VersionChangeDirective extends SubDirective
             // interlink-shortcode (from guides.xml) so it need not be repeated.
             $shortcode = $this->themeSettings->getSettings('interlink_shortcode');
             if ($shortcode === '') {
+                $this->logger->warning(
+                    'The ":changelog: #..." form requires "interlink-shortcode" to be set in the guides.xml. ',
+                    $blockContext->getLoggerInformation(),
+                );
+
                 return null;
             }
 
-            $target = str_replace('/', '-', $shortcode) . ':' . substr($changelog, 1);
+            $prefix = str_replace('/', '-', $shortcode) . ':';
+            $anchor = substr($changelog, 1);
         } elseif (str_contains($changelog, ':')) {
             // Explicit "<shortcode>:<anchor>". Normalize the shortcode to the
             // dash form the permalink service expects (vendor/package ->
             // vendor-package), exactly like the "copy permalink" button.
             $parts = explode(':', $changelog, 2);
-            $target = str_replace('/', '-', $parts[0]) . ':' . ($parts[1] ?? '');
+            $prefix = str_replace('/', '-', $parts[0]) . ':';
+            $anchor = $parts[1] ?? '';
         } else {
             // Bare value: a TYPO3 core changelog entry identifier.
-            $target = self::CHANGELOG_INVENTORY . ':' . $changelog;
+            $prefix = self::CHANGELOG_INVENTORY . ':';
+            $anchor = $changelog;
         }
 
-        return self::PERMALINK_BASE . $target;
+        if (trim($anchor) === '') {
+            $this->logger->warning(
+                'The ":changelog:" option has an empty changelog entry anchor. ',
+                $blockContext->getLoggerInformation(),
+            );
+
+            return null;
+        }
+
+        return self::PERMALINK_BASE . $prefix . $anchor;
     }
 }
--- a/packages/typo3-docs-theme/src/Directives/AbstractTypo3VersionChangeDirective.php
+++ b/packages/typo3-docs-theme/src/Directives/AbstractTypo3VersionChangeDirective.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace T3Docs\Typo3DocsTheme\Directives;
 
 use phpDocumentor\Guides\Nodes\CollectionNode;
+use phpDocumentor\Guides\Nodes\Inline\PlainTextInlineNode;
+use phpDocumentor\Guides\Nodes\Inline\ReferenceNode;
 use phpDocumentor\Guides\Nodes\Node;
 use phpDocumentor\Guides\RestructuredText\Directives\SubDirective;
 use phpDocumentor\Guides\RestructuredText\Parser\BlockContext;
@@ -17,7 +19,6 @@ use T3Docs\Typo3DocsTheme\Settings\Typo3DocsThemeSettings;
 use function array_values;
 use function explode;
 use function str_contains;
-use function str_replace;
 use function str_starts_with;
 use function substr;
 use function trim;
@@ -27,12 +28,12 @@ use function trim;
  * (versionadded, versionchanged, deprecated).
  *
  * All three directives support a ":changelog:" option that renders a link to
- * the related changelog entry, built as a docs.typo3.org permalink. The value
- * is not validated against an inventory, so a typo produces a link that 404s
- * when clicked.
+ * the related changelog entry. The value is resolved against the changelog
+ * inventory (interlink), so an entry that does not exist produces a warning and
+ * no link instead of a link that 404s when clicked.
  *
- * For a TYPO3 core change, pass the changelog entry identifier; it becomes a
- * "https://docs.typo3.org/permalink/changelog:<id>" link:
+ * For a TYPO3 core change, pass the changelog entry identifier; it is resolved
+ * against the "changelog" inventory:
  *
  * ..  versionchanged:: 14.0
  *     :changelog: feature-107628-1729026000
@@ -40,10 +41,9 @@ use function trim;
  *     This module has been moved from :guilabel:`System` to
  *     :guilabel:`Administration`.
  *
- * For an extension change, pass the extension's permalink shortcode followed by
- * the changelog entry anchor. The shortcode is normalized to the dash form the
- * permalink service expects ("vendor/package" -> "vendor-package"), just like
- * the "copy permalink" button on docs.typo3.org:
+ * For an extension change, pass the extension's interlink shortcode
+ * ("vendor/package") followed by the changelog entry anchor. It is resolved
+ * against that extension's inventory:
  *
  * ..  versionchanged:: 2.0
  *     :changelog: acme/acme-blog:changes-2-0-0
@@ -59,15 +59,11 @@ use function trim;
  *     :changelog: #changes-2-0-0
  *
  *     A configuration option was renamed; see the changelog for the migration.
- *
- * Note: the shortcode normalization cannot round-trip a vendor or package name
- * that itself contains a hyphen (a permalink-service limitation, see
- * https://github.com/TYPO3GmbH/site-intercept/issues/283).
  */
 abstract class AbstractTypo3VersionChangeDirective extends SubDirective
 {
-    private const PERMALINK_BASE = 'https://docs.typo3.org/permalink/';
     private const CHANGELOG_INVENTORY = 'changelog';
+    private const CHANGELOG_LINK_TEXT = 'See changelog entry';
 
     /** @param Rule<CollectionNode> $startingRule */
     public function __construct(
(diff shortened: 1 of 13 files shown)


linawolf:
The links should be resolved, for example https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/14.0/Feature-107628-ImprovedBackendModuleNamingAndStructure.html#feature-107628-1729026000 instead of https://docs.typo3.org/permalink/changelog:feature-107628-1729026000

And non existent changelog entries like "feature-101010-1700000000" should cause warnings and not be rendered to avoid links like https://docs.typo3.org/permalink/changelog:feature-101010-1700000000 to invalid shortcodes

CybotTM:
I also need feedback on the link rendering.

CybotTM:
@linawolf Done — `:changelog:` now resolves against the changelog inventory instead of building a `permalink/…` URL:

- a real entry renders the resolved link, e.g. `feature-107628-1729026000` → `…/Changelog/14.0/Feature-107628-…html#feature-107628-1729026000`
- a non-existent entry logs a build warning and renders no link (no more dead links to invalid shortcodes)

The rendered help page is in the PR description.

packages/typo3-docs-theme/src/Directives/AbstractTypo3VersionChangeDirective.php
```diff
@@ -0,0 +1,170 @@
...
+                    'The ":changelog: #..." form requires "interlink-shortcode" to be set in the guides.xml. ',
+                    $blockContext->getLoggerInformation(),
+                );
+
+                return null;
+            }
+
+            $interlinkDomain = '';
+            $anchor = trim(substr($changelog, 1));
+        } elseif (str_contains($changelog, ':')) {
```
linawolf: This re-implements domain:anchor splitting by hand (`str_contains`/`explode`) instead of reusing `ExtendedInterlinkParser`, which already parses this exact `domain:anchor` syntax for every other cross-reference in the project (e.g. via `GenericReferenceTextRole` for `:ref:`).

Two concerns with duplicating it:
- **Behavioral drift**: `ExtendedInterlinkParser` restricts the domain to `[a-zA-Z0-9\-_\/.]+` before the colon; this hand-rolled version doesn't enforce that, so a `:changelog:` value and an equivalent `:ref:` with the same text can be parsed differently.
- **Maintenance burden**: any future fix to the canonical parser (edge cases, new syntax) won't apply here, since this is now a second, independent implementation of the same logic.

Could `:changelog:` reuse `ExtendedInterlinkParser` (or a shared helper it exposes) instead of parsing the domain/anchor split itself?

CybotTM: You are right on all three counts, and I checked each rather than taking them as read.

`InterlinkParser::class` does resolve to `ExtendedInterlinkParser` (`typo3-docs-theme.php:148`), and `GenericReferenceTextRole`, `DocReferenceTextRole` and `ApiClassTextRole` all receive it — so it is the canonical split for every cross-reference here, and the hand-rolled one was a second implementation. The drift you predicted was already real: `other vendor:changes-2-4-0` was accepted as a shortcode named `other vendor`, where the parser's `[a-zA-Z0-9\-_\/.]+` rejects it.

Fixed in 14b1fc9: the option now delegates to the parser.

One thing worth flagging, because a naive delegation loses it. The parser returns an empty interlink for anything that is not a valid `<domain>:`, which covers the bare core-changelog form *and* a malformed one — so `:changelog: :anchor` would have silently become a local reference to a label named `":anchor"` instead of the warning it produces today. The empty result is therefore split on whether the value contains a colon at all: colon present means malformed and warns, no colon means the bare form.

That leaves one deliberate difference from `:ref:`, which I would rather state than have you find: after this change the *parsing* is identical, but the *outcome* for an unparseable value is not. `:ref:` treats it as a local label; `:changelog:` warns and renders nothing. For an option whose only job is to link a changelog entry, a silent local-label lookup seemed the worse failure. Say the word if you would rather have them identical.

Reviewing the rest of the branch off the back of this turned up three untested branches, all of them warnings in code this PR adds, now covered and each seen failing with its guard removed:

- an empty anchor, from `vendor/ext:` and from a bare `#`
- the two malformed inputs above
- the `#` form without `interlink-shortcode`, which is documented as warning but had no fixture — it needs its own manual, so it is a second fixture directory

Verified on the final tree: every gate the CI runs — `test-unit`, `test-integration`, `phpstan`, `test-xml`, `test-docs`, `test-monorepo`, `test-rendertest`, `composer normalize` — passes locally. `code-style` reports one file, `node_modules/flatted/php/flatted.php`, which is gitignored and only present because npm ran here; the quality job installs no `node_modules`, so CI never sees it. Worth excluding in `.php-cs-fixer.dist.php` at some point, but not in this branch.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_014H1xwaAmrQRWUA3vx8bJcD)_

linawolf:
Link building is fixed — the fixture shows a resolved URL, not a `permalink/…`
one. Building a `ReferenceNode` and letting the pipeline resolve it is the right
approach.

Two things, one question.

**The template can be one `renderNode`.** Today it renders the node, discards the
`<a>`, then hand-builds a second one:

```twig
{%- do renderNode(node.changelogReference) -%}
{%- if node.changelogReference.url %} <a class="versionchange-changelog" href="{{ node.changelogReference.url }}">…</a>{% endif -%}
```

It is correct — resolution happens in `ReferenceResolverPreRender`, so rendering
is what resolves — but `inline/link.html.twig` already emits `<a href>` with
`node.classesString` when `url` is set and plain children when it is not. Link
nodes inherit `setClasses()`, so set `versionchange-changelog` on the
`ReferenceNode` and this becomes:

```twig
{%- if node.changelogReference %}{{ renderNode(node.changelogReference) }}{% endif -%}
```

Same HTML, and no reading `url` from a template.

**Drop the `interlink-shortcode` requirement on the `#anchor` form.** The code
sets `$interlinkDomain = ''` for it — a local reference that resolves on its own
— but warns and bails when the setting is missing. That is a failure mode for a
link that would work, and it is the only reason the
`version-change-changelog-no-shortcode` fixture exists. Unless the shortcode
changes how the local form resolves and I have missed it, both can go.

**Unit tests?** `buildChangelogReference` is four branches and three warning
paths of pure string handling. Covering it directly would let the fixtures go
back to demonstrating rendering instead of enumerating malformed input. Happy
for this to follow separately.

CybotTM:
Both done, and the branch has moved a long way since; it is rebased on current `main` (`170119b4`).

**One `renderNode`.** The class sits on the `ReferenceNode` and the template is `{%- if node.changelogReference %} {{ renderNode(node.changelogReference) }}{% endif -%}`. `getChangelogLinkText()` is gone, and the link-text constant moved to the directive, its only remaining reader.

One thing your sketch changes, and I kept it rather than guard it away: this theme's `inline/link.html.twig` does not emit plain children in the `else` branch, it emits `<span class="invalid-link" title="This link cannot be resolved anymore">changelog:feature-101010-1700000000</span>`. So an entry that does not exist now shows that marker instead of nothing. It is still not a link, the warning is unchanged, and it is what every other unresolvable cross-reference in the project does. Say the word and I will put the `url` check back.

**`interlink-shortcode` on the `#anchor` form.** The setting never took part in resolving the local form. Precondition dropped, with the sentence in the docblock and on the rendertest page. It is still read for the one case that does need it, `<own-shortcode>:anchor` collapsing to a local reference.

I kept the second fixture and inverted it rather than deleting it: its `guides.xml` is the only one in the suite that sets no `interlink-shortcode`, so it now shows the local form resolving without one. It turned out to carry something else too — it is the only manual asserting that *no* warning log is written at all, and since every other expected log is compared by line containment, it is the only place a stray new warning cannot hide. That is now a note in the fixture. If you would still rather have it gone, it is one commit.

**Unit tests.** Agreed, as a follow-up: `buildChangelogReference` is private, so covering it directly means either extracting the parsing or driving the directive with a constructed `BlockContext`. I will open it separately once this is in, and move the malformed-input cases out of the fixture then. One branch is waiting for it — see below.

## A question for you, on the link text

The standardized "See changelog entry" wording has an accessibility consequence I would rather you decide than decide for you. On a page with several version-change blocks the links are identical in text and different in destination — measured on the fixture: 3 links, 1 link text, 2 distinct targets. WCAG 2.4.4 (A) is met, because the enclosing `<p class="versionmodified">` reads "Changed in version 2.1 See changelog entry" and the version disambiguates it in context. 2.4.9 (AAA) is not met, and in a screen reader's "list all links" view — a normal way to navigate — the entries are indistinguishable.

The awkward part is that the clean fix collides with the change you asked for. Giving each link a distinguishing accessible name (`aria-label="See changelog entry for version 2.1"`) means hand-building the `<a>` in the template again, which is exactly what the single `renderNode` replaced; and the link text originates as a `PlainTextInlineNode`, so neither an attribute nor a visually-hidden span can be injected without reverting that. The alternatives are to put the version into the visible text, which changes the standardized wording, or to accept AA and leave it. Your call — I did not want to quietly undo your review point to satisfy a AAA criterion.

Everything else on the accessibility side checks out: contrast 12.63:1, links are underlined so the link is not signalled by colour alone, and no new focus or keyboard behaviour — it is a plain `<a href>`.

## What the review rounds turned up

While this was waiting I put the branch through several review passes. Four defects in the `:changelog:` value path, and three of them are shapes of one mechanism:

- `:changelog:` written with no value is parsed as the boolean flag `true`, which stringifies to `"1"` — it rendered a `changelog:1` marker and logged a message naming neither the directive nor the option.
- The same option with its value on the following line arrives as `"1 feature-…"`, because the parser appends the continuation to that stringified flag.
- With a trailing space after `:changelog:` the prefix is a space instead of the `1` — and because the value was trimmed before it was examined, that form rendered a working link while the identical-looking form without the trailing space warned. Whether a link appeared was decided by an invisible character.
- The serious one: `preg_match` returns `false`, not `0`, on malformed UTF-8. An invalid byte therefore read as "no whitespace", reached `SluggerAnchorNormalizer`, and threw — ending the whole render, no pages written, with `Invalid UTF-8 string.` attributed to `body/version-change.html.twig`. A `.rst` saved as Latin-1 with `#änderung` in the value is enough.

The value is now read as an option rather than a string, untrimmed, with its encoding checked before the whitespace class. One branch has no fixture: the trailing-space form needs an `.rst` line ending in a space, and this repository's `.editorconfig` strips exactly that, so the case would be silently rewritten by any editor honouring it. It is verified by a rendered probe, noted in the fixture, and it is the first thing the unit tests should cover.

## Two things I did not change here

The self-shortcode comparison is case-sensitive, so `ACME/ACME-BLOG:x` attempts an inventory fetch — but `RemoveInterlinkSelfReferencesFromCrossReferenceNodeTransformer` does the identical case-sensitive compare, so normalising both is a decision about the project, not a drive-by in this PR.

And the integration runner compares warning logs by line containment. Extra warnings are therefore invisible in any fixture shipping an expected log, an expected log consisting of one blank line asserts nothing at all, and roughly three quarters of each expected HTML file lies outside the compared markers. That is `IntegrationTest` behaviour across all 119 data sets rather than anything this branch introduced — happy to open it as its own issue if you want it tracked.

Separately, #1409 fixes the unresolved-reference marker this PR now produces more often: its dotted underline never rendered, and its meaning was carried only by a `title` attribute.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_012Zm6UrYoeXRnucAdTdLRuN)_

linawolf:
Not AA versus AAA — I think the link text is the wrong text.

`See changelog entry` is the same string on every such link, which is what created the accessible-name problem in the first place. But the entry has a name of its own, and the pipeline already knows it. `InterlinkReferenceResolver` fills a reference with its target's title when the node has no children:

```php
if (count($node->getChildren()) === 0) {
    $node->addChildNode(new PlainTextInlineNode($link->getTitle()));
}
```

`AnchorReferenceResolver` does the same for the local `#anchor` form. The branch overrides that by passing `[new PlainTextInlineNode(self::CHANGELOG_LINK_TEXT)]`. Drop the children and the link reads:

> Changed in version 14.0 — Feature: #107628 - Improved backend module naming and structure

Which is self-describing, distinct per link, and dissolves the accessibility question entirely: no `aria-label`, no hand-built `<a>`, and the single `renderNode` stays. I checked the sizes against the published core changelog inventory — 3837 entries, median title 62 characters, shortest 26, longest 139. That carries as link text.

Take the full title including the `Feature: #107628 - ` prefix: it states the kind of change and the issue number, both worth having, and trimming it would be new special-case logic. Unresolvable entries keep the `invalid-link` marker with the raw shortcode, unchanged.

The four defects are a good catch, the UTF-8 one especially. `preg_match` returning `false` rather than `0` on malformed input, and that taking the whole render down with no pages written, is the kind of failure that only appears on somebody else's file at the worst moment.

I checked the branch against current `main` rather than the state CI ran on: merges cleanly, integration 128, unit 116, CS 0 of 164, PHPStan clean. So this is the only thing outstanding.

On the two you deliberately left alone, agreed on both. The case-sensitive self-shortcode comparison is a decision about the project and should not be a drive-by here. And yes, please open the `IntegrationTest` issue — part of it is already moving in #1420, where `.md` fixtures compare line by line, because dropping blank lines is right for HTML and wrong for Markdown, where a blank line separates paragraphs and closes code blocks. The warning-log containment and the unmatched region outside the markers are still open.

CybotTM:
You are right that the pipeline already knows the title, and I rendered your proposal against the live inventory rather than reasoning about it. It carries for the 3815 core entry labels. It does not carry in three places, and one of them produces a link with no accessible name at all — so I would like to propose a variant that keeps your mechanism as the default and closes those.

## What the three variants look like

Same probe manual, same renderer, `changelog` and the two extension inventories fetched live.

**Today — `See changelog entry`:**

Current link text

**Your proposal — drop the children, take the resolved title:**

Resolved title as link text

## Where it does not carry

**Sub-anchors.** 125 of the 3940 `Changelog/` labels are anchors inside an entry rather than the entry itself. `feature-93334-usage` resolves to the link text `Usage`, `changelog-howto-tagging` to `Tagging`.

**A label with no section parent renders an empty link.** `AnchorReferenceResolver` takes `$target->getTitle() ?? ''`, and `CollectLinkTargetsTransformer` gives an `AnchorNode` a title only when its parent in the shadow tree is a `SectionNode`. A label declared inside a directive body has none, so the `#anchor` form emits `<a href="#nested-label" class="versionchange-changelog"></a>` — a link with no accessible name, which is 2.4.4 at level A, while the wording question we started from is level AAA. This is inherited, not introduced: a bare <code>:ref:`nested-label`</code> renders the same empty `<a>` in the same run. But the branch as it stands does not have it, and that makes it a regression rather than a pre-existing quirk.

**Extensions, as extensions label their changelogs today.** Of eight inventories I checked: `georgringer/news` and `friendsoftypo3/tt-address` carry one label for the whole changelog page, so every link on a page reads `ChangeLog` or `Changelog` — identical on every link, which is the problem the change is meant to solve. `netresearch/nr-textdb` has a label per version, so `version-4-0-0` renders as `Version 4.0.0` directly after `Changed in version 4.0.0`. Three have no changelog label at all and two serve no inventory. The form the option was designed for is the third one, and there the title is usable but says what the badge already said. Extension docs are the reason this option exists, so this is the case I would most like to get right.

## What I would propose instead

Keep your mechanism as the default, put the link on its own line, and let the author override the text:

Title on its own line

Three tiers: author-supplied text if given, otherwise the resolved title as you propose, otherwise the fixed string when the title is empty. The line break is what makes a 139-character title tolerable — it no longer pushes the version label around, and the block reads as a heading, a link and a description rather than one run-on line.

Two costs, so they are on the table rather than discovered later. Author text needs a syntax, and the natural one is the form every other reference uses — `:changelog: Custom text <feature-107628-1729026000>` — which contains a space, so the single-token rule would have to become "a single token, or the `text <target>` shape". And the line break in the screenshot is a `<br>` for the probe; the real change is a class on the `<p>` and a rule in the theme.

There is also a cleaner place for the empty-title case than this PR: `AnchorReferenceResolver` falling back to the anchor name when the title is null would fix every bare `:ref:` in the project at once, and would let your version stand here with no fallback tier at all. If you prefer that, I am happy to open it upstream and rebase this on it.

If after this you would still rather have the bare title with no author override and no line break, say so and I will push it — it is one commit, plus whatever we agree on for the empty-title case.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01TdpwFRJQEjKJ5DHVWwTrH8)_

linawolf:
I measured the same inventory rather than reasoning about it, and two of the three do not hold at the size you gave.

**Empty titles: none.** Of 3940 `std:label` entries under `Changelog/`, zero have an empty title. The mechanism you describe is real, but the targets this option actually has do not hit it. That makes the level-A regression theoretical here — and it belongs in `AnchorReferenceResolver` anyway, exactly as you offered. Please do open that upstream; it fixes every bare `:ref:` and is the right place. It should not hold this PR.

**Sub-anchors: 39, not 125.** Splitting the 3940 by what the title actually is:

| | |
|---|---|
| entry titles | 3837 |
| sub-sections inside an entry | 39 |
| index and Howto pages | 64 |

The 39 sit in roughly 25 of 3402 entries. I suspect your 125 counted the index pages and the alternate entry anchors — `changelog-Feature-91008-…` and friends — but those resolve to the entry title, so they are not affected. And a sub-anchor only bites when an author deliberately targets a section instead of the entry; the text is then `Migration` rather than nothing.

**Extensions: confirmed, all three.** `georgringer/news` and `friendsoftypo3/tt-address` carry exactly one changelog label, so every link on a page reads `ChangeLog`. `netresearch/nr-textdb` has one per version, so `Version 4.0.0` lands directly after `Changed in version 4.0.0`. This is the one that stands, and as you say it is the case the option exists for.

**So: two tiers, not three.** Author-supplied text if given, otherwise the resolved title. No fixed fallback — it would only cover a case that does not occur, and the empty-title fix has a better home.

The author syntax earns its place on the extension evidence alone. `:changelog: Custom text <feature-107628-1729026000>` is the right shape; widening the rule to "a single token, or `text <target>`" is worth it for that.

On the line break: yes, but as its own thing. It is presentation, it is a theme change, and it should not ride along with a correctness fix.

CybotTM:
Built as you asked, rebased on current `main` (`e119a82d`), tip is `94dcb813`.

**Two tiers.** The reference is built without children, so the resolvers fill it with the entry's title; an author-supplied text takes that place. No fixed fallback, and the `See changelog entry` constant is gone.

**The embedded form.** `:changelog: Renaming the teaser field <acme/acme-blog:changelog>`, split by `EmbeddedReferenceParser` — the trait the reference text roles use — so the accepted shape cannot drift from theirs. Allowing a text means whitespace in the value is no longer a rejection on its own, so the two continued-value forms are now told apart by where the whitespace sits: a value the parser continued onto the option line begins with it and is rejected before the text is split off, and the entry is still required to be a single token once it is. Fixtures cover the embedded form over the core, extension and `#anchor` targets, brackets with no text before them, whitespace inside the brackets, and empty brackets.

`test-integration` 130, `test-unit`, `phpstan`, `code-style`, `test-xml`, `test-docs`, `test-rendertest`, `test-monorepo` all pass locally. The rendertest page shows both forms against a real core entry.

**Your sub-anchor count is the better one** — mine classified by label name, yours by what the title actually is, and the alternate anchors do resolve to the entry title. I will open the `AnchorReferenceResolver` fallback upstream separately.

**On the line break I am not convinced, and I would rather say so than quietly defer it.** Two things make it look like part of this feature rather than presentation on top of it: it targets `body/version-change.html.twig`, which this PR already changes, and `main` carries neither `changelogReference` nor `versionchange-changelog`, so there is no existing rendering it could alter — it can only ever shape markup this PR introduces. And it is the mitigation for the change you asked for: without it a 139-character title sits inline behind `Changed in version 13.4`, which is the state that lands if the follow-up does not. My preference is one commit for it in this PR, separate from the one above, so the history still separates them.

That is a preference, not a blocker. Say the word either way and I will follow it.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01TdpwFRJQEjKJ5DHVWwTrH8)_

linawolf:
You are right about the line break, and the manuals make the case better than either of us did. I counted the version-change directives in TYPO3 Explained — 240 of them:

| body | |
|---|---|
| prose of its own | 90 |
| contains a changelog permalink | 76 |
| empty | 42 |
| short prose | 27 |

Two things follow from those 76. Every one of them already writes the entry's title as the link text:

```rst
See `Deprecation: #108345 - Deprecation of ext_emconf.php <https://docs.typo3.org/permalink/changelog:deprecation-108345-…>`_
```

So the resolved title is not a new style — it is the one authors type by hand, and the option automates it. And that link lives in the body, under the badge, which is to say it already sits on its own line. An inline link would make the option look unlike the 76 places it exists to replace. (17 of the 76 are nothing but the link; the other 59 carry it alongside prose, often a list of several entries.)

Your other argument holds as well: `main` carries neither `changelogReference` nor `versionchange-changelog` — zero files, I checked — so the change cannot alter any existing rendering.

So yes: in this PR, as a separate commit, the way you proposed. My reason for deferring was that presentation should not ride along with a correctness fix. That assumes there is presentation to regress, and here there is none.

The rest reads right to me: one `renderNode`, the class on the node, and `EmbeddedReferenceParser` so the accepted shape cannot drift from the text roles.

CybotTM:
Done, as its own commit — `004427e0`.

The rule is one SCSS block, and it hangs on a wrapper rather than on the link's own class. An entry that does not resolve renders as the theme's `invalid-link` marker, which carries none of the reference node's classes; without the wrapper that one case would have been the only thing still sitting inline, next to a bold label. So `body/version-change.html.twig` wraps the reference in `<span class="versionchange-changelog-line">`, and the rule sets `display: block` and `font-weight: normal` on it — the second half because your 76 permalinks are not bold either.

Changelog link on its own line

The 139-character title is the second block; at 400 pixels it wraps to four lines and still leaves the version label alone:

The same at 400 pixels

`theme.css` is rebuilt through DDEV, so the built file moves by the twelve lines the rule generates and nothing else. Both changelog fixtures are rebaselined for the wrapper. `test-integration` 130, `test-unit`, `phpstan`, `code-style`, `test-xml`, `test-docs`, `test-rendertest`, `test-monorepo` all pass locally.

Thank you for counting the manuals — 76 of 240 writing the title by hand is a better argument than either of the two I had, and it settles the link text as well as the line.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01TdpwFRJQEjKJ5DHVWwTrH8)_

linawolf (approved):
Good work, thanks a lot!

linawolf:
The upstream PR is welcome — open it and I will review it there.

`AnchorReferenceResolver` fixes it for every bare `:ref:`, which is more
than this repository can reach from here.

Nothing here waits on it. From my side this is ready.

CybotTM:
Both are open.

Upstream: [phpDocumentor/guides#1395](https://github.com/phpDocumentor/guides/pull/1395), against `main`, with [issue #1394](https://github.com/phpDocumentor/guides/issues/1394) describing the case first, as their contribution chapter asks. The fix is one line — fall back to `$node->getTargetReference()` instead of `''` — plus two test cases: the titleless target that was empty before, and a titled one that guards against the fallback swallowing a real title. Against the unchanged resolver the first fails and the second passes, so the pair distinguishes the defect rather than just exercising the path. 979 tests, `code-style` and `phpstan` clean. No integration fixture changed, which also says none of them covered a titleless label until now.

Docs: [TYPO3CMS-Guide-HowToDocument#571](https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/pull/571), **draft on purpose** — the option is in render-guides `main` but not in `0.41.0`, which is what renders docs.typo3.org today. It documents the three value forms and the embedded `text <entry>` form, and every example is a `code-block` rather than a live directive, so the page renders unchanged on the current release and nothing needs revisiting when the option ships. `make test-docs` passes without warnings. Ready whenever a release carries it.

PR #1304, opened 2026-06-20
[TASK] Remove redundant bootstrap-reboot import

richardkrikler:
bootstrap.scss already includes root and reboot, so importing bootstrap-reboot separately causes both blocks to appear twice in the compiled CSS.

Removing it from theme.scss here has no side effects other than reduced theme.css filesize.


CybotTM (approved):
Verified as a rendering no-op:

- On `main`, the compiled `theme.css` contains the Reboot block twice, byte-identical except the heading selector-list order (a Sass `@extend` artifact, no cascade effect). This PR drops the first copy; the surviving one already won every cascade tie.
- `npx grunt sass` on this branch reproduces the committed `theme.css` byte-identically.
- Rendered `Documentation-rendertest` and compared before/after (6 pages × light/dark, full-page at 1440px): 0 differing pixels, and `getComputedStyle` of every element is value-identical.
- 504,372 → 491,425 bytes (−12.9 KB) — a real transfer saving, since cdn.typo3.com currently serves `theme.css` without content-encoding.

<details>
<summary>Before/after screenshots (right panel = pixel diff ×10; black = identical)</summary>

Index page, light mode — before vs. after vs. diff

Admonitions page, dark mode — before vs. after vs. diff

</details>

CybotTM:
Looks good to me.

PR #1305, opened 2026-06-21
[FEATURE] Open inline-code information-modal via dedicated button

richardkrikler:
[FEATURE] Open inline-code information-modal via dedicated button

This allows for normal selection of the inline-code without
opening the information-modal.

Using a HTML button together with aria-label allows for improved
accessibility.

Breaking: Changing to a separate button requires it to be visible
unconditionally for the modal to be openable.
Places where the "question"-icon was previously not displayed, will
now show it. The previous check for node.info.url got removed in order
to have an open-button present.

---

Can be tested/ verified on https://render-guides-rendertest.ddev.site/PhpInline/Index.html.

image
image
image

### DIFF
--- a/packages/typo3-docs-theme/assets/sass/components/_code.scss
+++ b/packages/typo3-docs-theme/assets/sass/components/_code.scss
@@ -115,9 +115,22 @@ a.code-block-edit {
     font-family: $font-family-monospace;
     border-radius: $border-radius;
     border: 2px solid $white;
-    background: $gray-100;
+    background-color: $gray-100;
     padding: 0.25em 0.5em;
     line-height: 27px;
+
+    &:has(button) {
+        padding-right: 0;
+    }
+
+    button {
+        background-color: $gray-100;
+        border-color: $gray-100;
+        margin-top: -0.2em;
+        padding: 0.2em;
+        min-height: unset !important;
+        box-shadow: unset !important;
+    }
 }
 
 .code-inline-long {
--- a/packages/typo3-docs-theme/resources/public/css/theme.css
+++ b/packages/typo3-docs-theme/resources/public/css/theme.css
@@ -24385,9 +24385,20 @@ a.code-block-edit .fa-check:before {
   font-family: "Source Code Pro", monospace;
   border-radius: 0.375rem;
   border: 2px solid #ffffff;
-  background: rgb(247.35, 247.35, 247.35);
+  background-color: rgb(247.35, 247.35, 247.35);
   padding: 0.25em 0.5em;
   line-height: 27px;
+}
+.code-inline:has(button) {
+  padding-right: 0;
+}
+.code-inline button {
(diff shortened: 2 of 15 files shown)


garvinhicking:
Yes!!!! Please, I also really don't like the auto opening modal.

I wonder if "?" Is the best icon for this. Maybe use something that indicates "code view" like:

</>
[]
(i)
(…)

or so? Because IMO this isn't really "help" but rather "details" or "info" or "more".

richardkrikler:
I'm not sure what might be best. Searching through the FontAwesome icon catalogue, I found the following somewhat fitting.

FA7 fa-code
SCR-20260621-qqfl
FA7 fa-info
image
FA7 circle-info with transparent background is "pro" only & filled circle-info is a little disturbing
SCR-20260621-qrhh

I would be in favor of fa-info.
What do you think @garvinhicking ?

CybotTM:
Is FA the default for TYPO3 docs? And if yes, is it intentional or "grown"?
Whatever, what I really have in mind: why not using inline SVG or Bootstrap icons. Inline SVG are the modern variant of handling icons AFAIK. And Bootstrap is also already loaded, isn't it?
From what I see, FA is only used for a hand full of icons, dunno if this satisfies the FA. But I may not see the full picture.

But this topic may be out of scope for this issue.

richardkrikler:
@CybotTM 

I see a lot more FA usage currently in this project → https://github.com/search?q=repo%3ATYPO3-Documentation%2Frender-guides+fa-+path%3A%2F%5Epackages%5C%2Ftypo3-docs-theme%5C%2F%2F&type=code&p=1
I don't see any inline SVGs here → https://github.com/search?q=repo%3ATYPO3-Documentation%2Frender-guides+svg+path%3A%2F%5Epackages%5C%2Ftypo3-docs-theme%5C%2F%2F&type=code

I also do not see the connection between bootstrap being loaded in this project and the wish for inline SVGs.
Using e.g. https://icons.getbootstrap.com/icons/info-circle/ would either require npm package `bootstrap-icons` being installed (then `<i class="bi bi-info-circle"></i>` can be used in the template), or the svg-snippet being placed in the twig template directly.

Pls correct me if I'm wrong.

(I also have no historical knowledge of this codebase...)
(I also think that completely switching to inline SVGs might be another issue.)

CybotTM:
> (I also think that completely switching to inline SVGs might be another issue.)

Yes, as I wrote, it is off-topic, sorry for the intercourse. Created #1306 for my question.

> I also do not see the connection between bootstrap being loaded in this project and the wish for inline SVGs.

There isn’t one, and I never claimed or implied anything of the sort.

linawolf:
I think we should postphone the question about fontawesome and now use it becuase it looks ok

linawolf (approved):
Very cool, I think everyone has been waiting for this change!!

PR #1308, opened 2026-06-23
Fix inotify extension installation in Dockerfile

himanshuramavat:
The original Dockerfile attempted to install the inotify extension via PECL, but the package is not available on PECL for the current PHP version (8.2). This caused the build to fail.

This change switches to using the system package manager (apt) to install the php${DDEV_PHP_VERSION}-inotify package, which provides the inotify extension for the specific PHP version in use.

This resolves the build failure and allows the Docker image to be built successfully.

The fix involved modifying `.ddev/web-build/Dockerfile.inotify` to replace:
```
RUN pecl install ${extension}
```
with:
```
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y php${DDEV_PHP_VERSION}-inotify
```

### DIFF
--- a/.ddev/web-build/Dockerfile.inotify
+++ b/.ddev/web-build/Dockerfile.inotify
@@ -1,7 +1,1 @@
-ENV extension=inotify
-SHELL ["/bin/bash", "-c"]
-# Install the needed development packages
-RUN (apt-get update || true) && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confnew" --no-install-recommends --no-install-suggests build-essential php-pear php${DDEV_PHP_VERSION}-dev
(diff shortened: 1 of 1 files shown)


jaapio (commented):
With the removal of the usage of pecl we can cleanup this script a bit. We do have now 2 times then apt update, and I think the build essentials are not needed anymore. 
Basically we can remove the whole line 4. 

It's a good practice to keep these files small.

himanshuramavat:
@jaapio Done, removed line 4 and the other PECL leftovers. The Dockerfile is now just the apt install for `php${DDEV_PHP_VERSION}-inotify`. Verified with `ddev restart --no-cache`; extension loads correctly.

jaapio (approved):

jaapio:
Thanks!

PR #1324, opened 2026-07-09
[BUGFIX] Regenerate code-php-short fixture for modal-button markup

CybotTM:
## Problem

Integration tests on `main` fail since 50146a4bfb22b05da02dcfc0da938e2d05cd2e47:

```
1) ...IntegrationTest::testHtmlIntegrationBetweenMarkers with data set "code-php-short"
Failed asserting that two strings are equal.
```

Every open PR inherits this failure via the merge ref (e.g. #1291, #1292).

## Cause

#1294 added the `code-php-short` test with expected output generated before #1305 changed `code.html.twig` (information modal now opened via a dedicated button; `data-morelink` only emitted for a non-empty URL). Both PRs were green individually and merged past each other today.

## Fix

Regenerate the fixture via `make integration-baseline` — only `tests/Integration/tests/code/code-php-short/expected/index.html` changes. The `php-short` rendering itself is unaffected: names remain stripped of the leading backslash (`FooBar`, `Something`).

Integration suite local result: `Tests: 115, Assertions: 1150` — no failures.

PR #1326, opened 2026-07-14
[TASK] Add merge_group trigger for merge queue

CybotTM:
## What

Adds the `merge_group` event trigger to the Main workflow and documents that its job names are the merge queue's required-check contexts.

## Why

Follow-up to the team decision to trial a merge queue on this repository (and only this one). Background: #1294 was merged with checks that predated #1305's template change, breaking `main` until #1324 — the class of failure a merge queue prevents by re-running checks against current `main` before each merge.

Without this trigger, checks never run for queued merges and every merge group times out, so this must land **before** the queue is enabled.

## Next steps (after this PR merges)

1. Branch ruleset on `main`: merge queue (squash, all-entries-green) + the seven Main checks as required status checks; classic "require branches to be up to date" flag dropped as redundant.
2. Trial with a real PR and a Dependabot PR, results reported back to the team.

A decision record for the trial (scope: this repo only, not org-wide) is proposed in TYPO3-Documentation/.github.

## Note for #1196

The reusable-workflows migration rewrites this workflow's `on:` block and renames the reported check contexts. It must keep the `merge_group` trigger and update the ruleset's required-check contexts in the same merge window, otherwise the queue blocks all merges.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -5,11 +5,18 @@ on: # yamllint disable-line rule:truthy
   push:
     branches:
       - "main"
+  # Required for the merge queue on "main": without this trigger, queued
+  # merges never receive their required status checks and time out.
+  merge_group: null
 
 env:
   DEFAULT_PHP_VERSION: "8.2"
   RUN_ENVIRONMENT: "local"
 
+# The job names below ("Tests (8.x)", "Quality", ...) are configured as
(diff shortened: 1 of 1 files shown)


garvinhicking (approved):
Lgtm to test this out and gather experience.

garvinhicking (approved):

Issue #1344, opened 2026-07-25
Info modal renders helpText markup and pre-encoded entities as literal text


CybotTM:
Since ab24ff76 the code info modal shows markup and HTML entities as visible text. Two separate causes, both in this repo — no documentation source is involved.

## Where the content comes from

`data-details` is `CodeInlineNode::getHelpText()`. For `:php:` / `:php-short:` roles it is assembled in `packages/typo3-docs-theme/src/TextRoles/PhpTextRole.php`:

```php
$infoArray[] = '<code>' . implode(' ', $modifiers) . ' ' . $apiInfo['short'] . '</code>';
$infoArray[] = '<em>' . $apiInfo['summary'] . '</em>';
$info = implode('<br>', $infoArray);
$info = str_replace('\\', '&#8203;\\', $info);
```

`$apiInfo['summary']` comes from `https://api.typo3.org/<version>/api-info.json`.

## 1. Markup is rendered as text

`packages/typo3-docs-theme/assets/js/code-inline.js:74`

```js
content.innerHTML += `<p>${htmlEscape(item.dataset.details)}</p>`;
```

Together with the `|raw` removal in `code.html.twig`, the assembled HTML now reaches the user as text. The modal for `\TYPO3\CMS\Frontend\Event\ModifyPageLinkConfigurationEvent` displays:

```
<code>final class ModifyPageLinkConfigurationEvent</code><br><em>A generic PSR 14 Event …</em>
```

Affects every `:php:` / `:php-short:` role that resolves against the API.

## 2. Entities from the API are double encoded

`api-info.json` ships summaries that are already HTML encoded — **440 of 3803 entries**, e.g.

```
\TYPO3\CMS\Frontend\Event\ModifyPageLinkConfigurationEvent
  "A generic PSR 14 Event to allow modifying the incoming (and resolved) page when building a &quot;page link&quot;."
\TYPO3\CMS\Backend\Authentication\Event\SwitchUserEvent
  "This event is triggered when a &quot;SU&quot; (switch user) action has been triggered"
```

`PhpTextRole:154` takes them verbatim, Twig escapes the `&` to `&amp;quot;`, `htmlEscape()` escapes once more, and `&quot;` becomes visible.

Before ab24ff76 this was invisible rather than correct: reading the attribute through `element.dataset` resolved the entity exactly once, so the raw `innerHTML` happened to render `"`.

## 3. Unverified, same mechanism

[...]

PR #1345, opened 2026-07-25
[BUGFIX] Render class info modal from plain-text data attributes

CybotTM:
Proposal for option **a** in #1344 — something concrete to compare against the sanitizer route. Merge only if that is the option you want.

## What it does

`PhpTextRole::getClassCodeNode()` no longer assembles HTML. Signature, flags and summary travel as separate plain-text `data-*` values, and the modal writes the markup around them:

```js
if (item.dataset.signature) {
  detailParts.push(`<code>${htmlEscape(item.dataset.signature)}</code>`);
}
```

Every value is escaped exactly once, and no markup crosses the attribute boundary — nothing to sanitize, no allowlist to maintain.

The API summary is decoded on read, so plain text is what travels internally:

```php
$apiInfo['summary'] = html_entity_decode($apiInfo['summary'], ENT_QUOTES | ENT_HTML5, 'UTF-8');
```

Rendered result for the example from the issue:

```
data-signature="final class ModifyPageLinkConfigurationEvent"
data-summary="A generic PSR 14 Event … when building a &quot;page link&quot;."
```

The `&quot;` there is Twig escaping one real `"` for the attribute; `element.dataset` hands the modal a plain quote.

## Two deliberate behaviour changes

- The details block for `:php-short:` loses its second FQN line. `data-fqn` already carries the FQN and the modal renders it in its own section.
- The zero-width-space entity that `str_replace('\\', '&#8203;\\', …)` inserted around backslashes goes with the HTML blob. It only ever applied to the details string.

Both are visible in the diff and are why this is a proposal rather than something I would merge on my own.

## Verification

```
npm test                          6 files, 27 tests passed
make test-integration ENV=local   Tests: 115, Assertions: 1150, Skipped: 2   OK
make test-unit ENV=local          OK (83 tests, 183 assertions)
make phpstan ENV=local            No errors
make code-style ENV=local         0 of 147 files
```

PHP 8.5.8. One integration fixture changes, the changelog one; the `typo3-file` fixture is a non-API role and keeps the escaped plain text from #1343.

`ModalDetails.test.js` covers what the modal renders, which is the gap #1344 describes: signature and summary become `<code>` / `<em>` elements, a quote in the summary stays a quote, plain `data-details` still works for roles without API info, and markup arriving in either field stays inert. The XSS regression tests from ab24ff76 pass unchanged.


linawolf:
I think transporting the plain text values is much cleaner. Does this work for all kind of supported languages or only for PHP?

CybotTM:
Hi @linawolf ,

> Does this work for all kind of supported languages or only for PHP?

works for all.
nothing in the template or the modal JS is PHP-specific. The other roles just have no metadata to put in signature / flags / summary yet, only a static sentence.

PR #1364, opened 2026-07-31
Ineswillenbrock patch 1

ineswillenbrock:


CybotTM (approved):
looks good to me - only raises lower boundary, has no effect on actual installed packages.

PR #1374, opened 2026-08-14
[TASK] Rename Sitemap/Index.rst to Sitemap.rst

richardkrikler:
The theme only renders the special sitemap page tree when the source file is named exactly "Sitemap". Placing it at Sitemap/Index.rst, made the filename "Index" instead, so the page rendered empty.
Copied content over from
TYPO3CMS-Reference-Typoscript/blob/main/Documentation/Sitemap.rst.


sarahmccarthy123 (commented):
Looks good :-)

linawolf (approved):

PR #1388, opened 2026-08-25
[TASK] Add some basic stylings

benjaminkott:


linawolf:
**`_propertyCard.scss` — `.property-card > dt` is missing `display: flex`**

The new `align-items: center` / `column-gap` on `& > dt` are added, but the old `display: block` was removed without a replacement — so they're currently inert. Not a visible regression (main already had `display: block`), but the flex-row header this seems to be going for never actually applies.

Would be visible on `dl.command > dt` (simple name + headerlink) if fixed. `dl.confval` already works around it via its own hardcoded `d-flex`. Careful with `dl.php > dt` though — it often has many inline children (name, parens, multiple params, return type), so a blanket `display: flex` there would insert `column-gap` between every token and break the signature spacing. May need a narrower fix than just adding `display: flex` to the shared rule.

linawolf (approved):

PR #1273, opened 2026-06-01
[TASK] Improve responsive header layout and hide breadcrumb on root pages

hahnmg:
## Summary

- Refactors the page header markup into semantic wrapper elements (`header-nav`, `header-actions`) for cleaner, more maintainable responsive styling
- Fixes header layout on mobile: project title wraps correctly, search expands as an overlay on small screens, options panel is hidden on mobile
- Hides the breadcrumb `<ol>` on root/level-1 pages — only pages with at least one titled ancestor show the navigation trail

PR #1391, opened 2026-08-28
[BUGFIX] Fix partially broken open-search button on mobile

richardkrikler:
Clicking the top-right search icon on mobile revealed a panel whose only content was *another* "Search" button - a second tap was needed to actually reach the search dialog. Sometimes the triggering did not open the second button at all.

https://github.com/user-attachments/assets/57b808c8-15a3-4a81-bc54-4a6e72e23cc6

This PR consists of two commits:
1. fixes the bug
2. general cleanup

Instead of displaying a second button, the mobile button directly calls (clicks) the button coming from the React search component & triggering opening the search-dialog.
Probably a regression of: #1273

The second commit removes the hidden and unused HTML form element & associated stylings.

Tested also via connection to local t3docs-search-indexer instance & browser-stack iOS simulation.


linawolf:
Thanks for your work. I merged it now. A test #search-toggle triggers the modal would stop this regressing a second time would be great.

I rebased your branch and rebuild the assets.

linawolf (approved):

richardkrikler:
@linawolf Thank you!
Test for the two buttons => https://github.com/TYPO3-Documentation/render-guides/pull/1411

Issue #1400, opened 2026-09-10
[FEATURE] Render every page to Markdown alongside HTML (rel="alternate" / rel="canonical")


linawolf:
## What

Render a Markdown representation of every page next to its HTML, and point at it
from the HTML `<head>` — the way Apple does it in their developer documentation:

```html
<link rel="alternate" type="text/markdown" href="https://developer.apple.com/documentation/safari-developer-tools/enabling-developer-features.md">
<link rel="canonical" href="https://developer.apple.com/documentation/safari-developer-tools/enabling-developer-features">
```

So `…/Feature.html` would be accompanied by `…/Feature.md`, and the HTML page
would declare both the canonical URL and the Markdown alternate.

## Why

Tools that read our documentation programmatically — AI assistants above all —
want the content without the surrounding HTML. Today the only machine-friendly
form we offer is the reStructuredText source, and that is not good enough:

- includes, substitutions and interlink references are **not resolved** in the
  source, so a consumer sees `.. include::` instead of the content;
- the source is not what the page actually says — directives, confvals and
  tabs only become content during rendering.

This is not a source export and not a conversion of the generated HTML. It is a
**render target on the same footing as HTML**: the same AST, the same rendering
pipeline, its own set of templates. Every node type gets a deliberate Markdown
rendering that reproduces what the HTML page shows — the expected output of the
page, written in Markdown.

The `rel="canonical"` half is worth having on its own, independently of
Markdown: we currently emit **no canonical link at all**
(`resources/template/structure/layoutParts/generalHeaderLinks.html.twig`),
even though the same manual is reachable under several version URLs.

## What already exists

Most of the machinery is in place, which is what makes this attractive:

- **Multiple output formats are normal here.** `guides.xml` already renders
  `html`, `singlepage` and `interlink`, and `Documentation-markdowntest/`
  renders with `<output-format>rst</output-format>`.
- **A renderer that writes a non-HTML format from the AST already ships**:
  `phpdocumentor/guides-theme-rst` is a set of `*.rst.twig` templates plus a
  `TypeRenderer` tagged `'format' => 'rst'`. A Markdown theme is the same shape
  with `*.md.twig` templates.
- **The file extension comes from the format name for free.** `RenderContext`
  builds the output path as `$documentNode->getFilePath() . '.' . $outputFormat`,
  so a format named `md` writes `Feature.md` beside `Feature.html` with no
  path handling of its own — exactly the Apple layout.

## Proposed scope

[...]

PR #1401, opened 2026-09-10
[FEATURE] Link the reStructuredText source from the HTML head

linawolf:
[FEATURE] Link the reStructuredText source from the HTML head

Tools that read our documentation programmatically - AI assistants above
all - want the content without the surrounding HTML. Until every page is
rendered to Markdown as well, the reStructuredText source we already copy
into `_sources/` is the only machine-friendly form we offer, and nothing
in the page points at it: it is reachable only through the "View source"
entry in the options menu.

Emit it as an alternate representation in the `<head>`:

    <link href="_sources/index.rst.txt" rel="alternate"
          title="reStructuredText source" type="text/x-rst">

The link reuses the `copy_sources` setting and the same `copyDownload()`
call as the "View source" button, so it appears exactly where the `.txt`
file is actually written. A single-page render has no single source file,
so `sourceFilename()` is empty there and no link is emitted.

`text/x-rst` is the type docutils and PyPI use for reStructuredText;
there is no IANA-registered type for it.

Part of: #1400


linawolf:
We wont do it, the Rest is missing to much information to really recommend it instead we will render to markdown

Issue #1306, opened 2026-06-21
Drop Font Awesome in favor of inline SVG or Bootstrap icons


CybotTM:
The theme currently depends on Font Awesome and ships icon fonts.

I consider icon fonts legacy technology. For modern documentation UI, inline SVGs or a small local SVG sprite would be cleaner:

- no webfont loading
- smaller asset surface for a small icon set
- better accessibility control
- less dependency overhead

Since Bootstrap is already used by the theme, Bootstrap Icons could also be considered.

Any opinions on this?

linawolf:
Fontawesome icons could also be loaded as svgs instead of webfont if desired. I would like to stay with fontawesome as there is usually an icon for what you need and bootstrap has discontinued their icons in the past more then once

CybotTM:
Hi @linawolf 

So Bootstrap Icons would be a no for you, but switching Font Awesome from webfonts to SVGs would still be something you would support?

If so, I would keep this issue open but narrow its scope to replacing the Font Awesome webfont usage with local SVGs.

Otherwise we should close this issue.

btw.

> bootstrap has discontinued their icons in the past more then once

only the glyph fonts, in favor for SVG and a separate project AFAIK

CybotTM:
An accessibility note for whatever this issue lands on — with a correction to what I first wrote here, because my premise was wrong.

**What I got wrong.** I reported that of the Font Awesome icons in the theme templates, nearly all lack `aria-hidden="true"` and are therefore announced by screen readers as noise. The count is right — **49 of 52 icon elements across 21 templates** carry no `aria-hidden`, the only three that do are in `structure/layoutParts/pageHeader.html.twig` (lines 23, 27, 43) — but the consequence I drew from it is not. This theme resolves `@fortawesome/fontawesome-free` to **7.3.1** (`^7.3.1` in `package.json`, held there by the lockfile), and version 7 changed the default. Its webfont CSS gives every icon *style* class `content: var(--fa)/""` — an empty CSS *alternative text* — so in a browser supporting that syntax these icons already expose an empty accessible name and are ignored by assistive technology. Font Awesome's own documentation for v7 says as much. My "noise with no information in it" was true of v5 and v6, not of the version actually installed here.

[...]

PR #1409, opened 2026-09-12
[BUGFIX] Make the unresolved-reference marker perceivable

CybotTM:
## What

A reference that cannot be resolved renders as `<span class="invalid-link" title="This link cannot be resolved">…</span>`. After this change the marker actually carries a dotted underline, so it is distinguishable from surrounding text by something other than colour. A dotted *decoration* was clearly intended — `dotted` is meaningful only as a decoration style — though the declaration that introduced it never said which line.

## Why

`text-decoration: dotted` sets only the *style*. The shorthand is `<'text-decoration-line'> || <'text-decoration-thickness'> || <'text-decoration-style'> || <'text-decoration-color'>`, and an omitted longhand resets to its initial value — `text-decoration-line: none`. So nothing was ever drawn. Measured with `getComputedStyle` on `span.invalid-link`:

| | before | after |
|---|---|---|
| `textDecorationLine` | `none` | `underline` |
| `textDecorationStyle` | `dotted` | `dotted` |
| `textDecorationColor` | `rgb(51,51,51)` | `rgb(51,51,51)` |

That left the marker distinguished from surrounding text by a `#f2f2f2` background and nothing else — same colour, same font, no decoration, no pseudo-element content. **WCAG 2.2 § 1.4.1 Use of Color (Level A)**: *"Color is not used as the only visual means of conveying information … or distinguishing a visual element."* Failing before, passing after. The decoration colour equals the text colour, so the new cue is achromatic.

Two smaller things in the same element:

- The wording loses "anymore". It asserts that the reference once worked; the fixture's two references never resolved, and the build logs `Inventory link … not found` for both.
- A dead branch goes: the span sits inside the `{% else %}` of `{% if node.url %}` and then opened `{% if node.url %}` again, so its first arm could never be taken. `node.url` is a plain getter over a `private string $url = ''`, no side effects — the arm was unreachable.

## What this deliberately does not do

An earlier revision added the sentence as visually-hidden text so assistive technology would encounter it. That is withdrawn: `confval-menu` renders a field's content and then runs `striptags` over the result, so with `max=` set the sentence became visible table text and added 39 characters to the length that decides whether the value gets truncated — and, where that length then exceeded `max`, became the cell's tooltip too. Keeping it would have meant coupling two templates through a literal string, or adding a regex filter to an unrelated feature.

So the meaning still lives only in the `title` attribute. Chrome exposes that as the marker's accessible description, but a sighted keyboard or touch user sees the underline and can reach the explanation only with a mouse. Worth solving — it needs a mechanism that does not add a text node, and that is a decision for this theme rather than a detail of this fix.

## Testing done

`tests/Integration/tests/interlink/interlink-warn` is the only fixture rendering an unresolvable reference — verified by a repo-wide search of `expected/` directories, excluding generated `temp/` output — so it is the only expected file that changes.

Local run of every gate the CI jobs execute — `test-unit`, `test-integration`, `phpstan`, `test-xml`, `test-docs`, `test-rendertest`, `test-monorepo`, `composer normalize`, `npm test` — all pass. `code-style` reports only `node_modules/flatted/php/flatted.php`, which is gitignored and absent in the quality job.

The built `theme.css` is regenerated with `npm run build`; a no-op rebuild on `main` was confirmed byte-identical first, so the only CSS change in the diff is the one line. Note that no CI job runs `npm run build`, so the committed asset is not gated — it was checked locally instead.

## Not in scope

Decorative Font Awesome icons carry no `aria-hidden="true"` — 49 of 52 icon elements across 21 templates. That is unrelated to this marker, so it is noted on #1306 rather than bundled in here. It is also less urgent than it first looks: the theme resolves Font Awesome to 7.3.1 (`^7.3.1`, held by the lockfile), and its webfont CSS gives every icon style class `content: var(--fa)/""`, an empty alternative text, so these icons already expose an empty accessible name where that syntax is supported. `aria-hidden` remains worth adding for the `@supports not` fallback branch, and a migration to hand-inlined SVG would remove the empty name again — which is the part worth deciding explicitly over there.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_012Zm6UrYoeXRnucAdTdLRuN)_


linawolf:
That is a good idea, ping me when you are done

CybotTM:
@linawolf Done — out of draft and ready for you.

It ended up smaller than it started. An earlier revision also added the sentence as visually-hidden text so screen readers would meet it; that is withdrawn, because `confval-menu` renders a field's content and then runs `striptags` over it, so with `max=` set the sentence became visible table text and added 39 characters to the length that decides whether the value gets truncated. Keeping it would have meant coupling two templates through a literal string.

So what is left is three small things in one element: the dotted decoration actually renders (`text-decoration: dotted` sets only the style, and the shorthand resolved the line to `none`, so nothing was ever drawn — measured `textDecorationLine: none` before, `underline` after); the wording loses "anymore", which asserted that the reference once worked, while the fixture's two references never resolved; and a dead `{% if node.url %}` arm nested inside its own `{% else %}` is removed.

One thing I decided not to decide for you: the meaning of the marker still lives only in the `title` attribute. A sighted keyboard or touch user sees the underline and can reach the explanation only with a mouse. Fixing that needs a mechanism that adds no text node — which is a choice about this theme rather than a detail of this fix, so I left it.

I also corrected my own note on #1306 substantially: I had claimed those icons are announced as noise, which was true of Font Awesome 5 and 6 but not of the 7.3.1 this theme resolves to — v7's webfont CSS gives them an empty alternative text. The part that survives is smaller, and it turned up something separate worth your eye: five copy buttons in `linkReferenceModal.html.twig` consist of an icon alone with no `aria-label` or `title`, so they currently have no accessible name at all.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_012Zm6UrYoeXRnucAdTdLRuN)_

packages/typo3-docs-theme/resources/template/inline/link.html.twig
```diff
@@ -6,11 +6,7 @@
     {{- node.value -}}
     </a>
 {%- else -%}
-    <span class="invalid-link" title="This link cannot be resolved anymore">
-        {%- if node.url -%}
```
linawolf, line 10: why did you remove the value?

linawolf: Ah, nested in its own else

linawolf (approved):

PR #1410, opened 2026-09-12
[FEATURE] Render every page to Markdown alongside HTML

linawolf:
Part of #1400.

Tools that read our documentation programmatically want the content without the
surrounding HTML. The reStructuredText source is not good enough: includes,
substitutions and interlinks are unresolved there, and directives only become
content during rendering.

A new package `typo3-docs-theme-md` registers an `md` output format on the same
AST and pipeline as HTML, so `RenderContext` writes `Feature.md` beside
`Feature.html`. The theme ships it; `render-markdown="false"` switches it off.

Markdown is GitHub Flavored — admonitions and version changes become alerts,
tables become pipe tables. Links are permalinks built from
`interlink_shortcode`, because a downloaded file has to keep working.

**Verified against 39 real manuals**: the core changelog, the TCA, TypoScript,
ViewHelper and exception references, eleven guides and tutorials including
TYPO3 Explained, the 21 system extensions that ship their own documentation,
and the news and content-blocks extensions as third-party cases whose
interlink shortcode is a composer name. 10264 Markdown files.

- No unrendered directive anywhere. A node without a template leaves an HTML
  comment naming it; the rendertest started at 243 of those, and every manual
  now reports none.
- Every file opens with front matter that parses as YAML, and every manual has
  exactly one page marked as its start.
- Every permalink into a manual's own inventory resolves, and carries the
  version wherever the manual declares one: all 92 in the TCA reference and 382
  of the 383 across the system extensions, none of the 3912 in the changelog,
  which is published once.
- No HTML entities beyond the ones the manuals document on purpose — pages
  about `htmlentities()`, `htmlparser` or XML examples, where HTML shows the
  same characters.
- Nothing the HTML carries that the Markdown lacks. Comparing the text of both
  page by page, what is left is the HTML splitting identifiers across
  syntax-highlighting spans, and controls: "Copied!", "Edit on GitHub", the
  breadcrumb.

Rendering Markdown alongside HTML costs about 2%, which is less than the spread
between two identical runs — 126s to 142s for the same changelog — so the real
figure sits inside the noise.

Fifteen commits. The first three are the render target, the setting with its
download button and the `rel="alternate"` link, and the escaping together with
confval, rubric and the text roles. The rest add one directive family each, in
the order the manuals turned them up, and close with the front matter that
tells a file which manual, version and source it came from. Four are bugfixes:
an image lost its alt text, the file text role lost its code formatting, a code
block lost the caption that says which file the snippet belongs in, and the
permalinks named no version, so every link in a 13.4 manual led to whatever is
current later.

This supersedes #1401, which added the same `rel="alternate"` link pointing at
the copied reStructuredText source.

Signed-off-by: lina.wolf
Assisted-by: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_019EuSAzF5qw81NzPusWy2hC


benjaminkott (approved):
looks fine for me, i think we should take it as is and fix eventually open tings as we discover them

PR #1411, opened 2026-09-12
[TASK] Add regression test for the mobile search-toggle fix

richardkrikler:
#1391 fixed #search-toggle opening a dead-end panel instead of the search dialog. Add a guard so it can't silently regress a second time: covers both entry points into SearchModal, the mobile #search-toggle (forwarded via menu-expandable.js) and GlobalSearch's own desktop button, asserting each opens the dialog.


linawolf (approved):
Cool! Thanks

PR #1436, opened 2026-09-17
[BUGFIX] Render a table without a single cell as nothing, and size the ones that have cells by their widest row

### LINKED ISSUE #1435: Exception on MD rendering
I get an error when rendering the docs of my TYPO3 extensions sf_event_mgt since Version 0.41.0 of render-guides.

**Output:**

```
Rendering: 98/98 [============================] 100% Output format html: Rendering: //ForAdministrators/Configuration/TsReference/Index.md
Rendering: 98/98 [============================] 100% Output format singlepage: Rendering: //ForAdministrators/Configuration/TsReference/Index.md
Rendering: 98/98 [============================] 100% Output format interlink: Rendering: //ForAdministrators/Configuration/TsReference/Index.md
Rendering: 29/69 [===========>----------------]  42% Output format md: Rendering: //ForAdministrators/Configuration/TsReference/Index.md

In table.md.twig line 1:
                                                                                                                                                                              
  An exception has been thrown during the rendering of a template ("implode(): Argument #1 ($array) must be of type array, string given") in "body/table.md.twig" at line 1.  
                                                                                                                                                                              

In MdExtension.php line 162:
                                                                       
  implode(): Argument #1 ($array) must be of type array, string given                                                              
```

CybotTM:
Rendering a manual to Markdown no longer aborts when a table has no cell in it at all, and a table whose header is narrower than its body no longer loses the cells beyond the header. A `t3-field-list-table` with nothing indented under it now warns, naming the file and the line.

## The crash

`renderMdTable()` fills in an empty header row when the source table has only body rows, but the fill is guarded by `$columnCount > 0` while the line after it reads `$rows[0]` unconditionally. With neither header nor data, `$rows[0]` is undefined and `implode()` gets `null`. On PHP 8.2 that lands on `implode()`'s legacy single-argument signature, which is where the reporter's wording — `Argument #1 ($array) must be of type array, string given` — comes from; on 8.5 the same call says `null given`.

GFM has no way to spell a table with no cell, not even the delimiter row, so the guard now returns an empty string instead. The HTML is unchanged and still renders the empty table it always did.

## Two more degenerate tables, found in review

The header row and the delimiter row were sized by `count($rows[0])`, the width of the *first header row*, while each body row was emitted at its own width. Per the GFM spec a body row's excess cells are ignored, so a header naming one field and rows carrying two rendered as

```
| Header1 |
| --- |
| 1 | one |
```

and `one` is not shown by any Markdown reader; the HTML shows it. A header row with *zero* columns — reachable whenever the first list item is not a field list, which already logs a warning — produced a delimiter row with no dash in it, which is not a delimiter row at all, so the whole block displayed as literal pipes. Both rows are now sized by the `$columnCount` the function already computes.

## Header rows GFM cannot hold

GFM knows exactly one header row, and only `$rows[0]` was ever written. A grid table with two rows above its `+===+` line, or a `list-table` with `:header-rows: 2`, therefore lost its second header row from the Markdown while the HTML shows it. Every header row after the first is now written as a body row, and every row is written at the width of the table: GFM fills missing cells in itself, a reader that does not (Python-Markdown's `tables` among them) would otherwise show a short row under the wrong columns.

## The directive says when it is empty

`t3-field-list-table` with no content produced an empty table and said nothing: silence in the HTML, and since the fix above, silence in the Markdown. It now logs a warning with `rst-file` and `currentLineNumber`, the same way the directive's two existing warnings do. The condition is "the directive has no child at all", not "no row could be built": content that is present but unusable — a paragraph where a list belongs — is already reported by the warning above it, and must not be reported twice (measured on a probe: two warnings before, one after). Manuals rendered with `--fail-on-log` will see this as a new failure — that is the intent; the directive was already producing nothing. In the reporter's manual exactly 1 of its 33 `t3-field-list-table` directives is affected. The directive's neighbouring "Only field lists are allowed" warning formatted the enclosing list instead of the offending child, so it always said `ListNode`; it now names what it found.

## Verification

Rendering `derhansen/sf_event_mgt` — whose TypoScript reference opens with a `t3-field-list-table` that has no content under it — reproduces the crash on `main` at page 29 of 69 and completes all 69 pages with this change, the confvals of that page intact.

New integration case `tests/Integration/tests/markdown-rendering/table/` covers each direction: a filled `t3-field-list-table` and a grid table still render as pipe tables, a header narrower than its body is widened so no cell is lost, the reporter's empty directive renders nothing while the prose around it survives, and `expected/logs/warning.log` pins the new warning. The expected files are generated, not hand-written.

The case grew with each review round and now also covers a table with two header rows and a list item that is not a field list. Every change was mutation-tested against it: removing the empty-table guard fails with the original `TypeError`, removing the header padding fails on the rendered table, removing the warning fails on the missing `warning.log`, dropping the extra header rows again fails on the rendered table, and restoring the old node name in the warning fails on `warning.log`. So each assertion is wired to a defect rather than merely green.

One thing the suite cannot pin: `assertFileContainsLines` checks that the expected log lines are *contained* in the actual log, so a surplus warning line never fails a fixture. That the empty-content warning no longer fires twice is therefore measured on a probe (two warnings before, one after), not asserted. Changing that assertion is a change to the shared harness and belongs in its own pull request.

One finding from the review is filed rather than fixed: `t3-field-list-table` never reads its `:header-rows:` option — [#1437](https://github.com/TYPO3-Documentation/render-guides/issues/1437). It predates this branch, affects HTML the same way, and what the directive should do with a value other than 0 or 1 is a decision. The new fixture pins today's behaviour, so that expectation moves with the fix.

`make pre-commit-test ENV=local` passes: 132 integration tests run and 2 skipped, 127 unit tests, php-cs-fixer, phpstan and monorepo-builder clean. The reporter's manual renders end to end from its extension root — 69 pages in all five formats (html, singlepage, interlink, md, tocjson), exit 0, no warning and no error log.

## Note for the reporter

The directive in `ForAdministrators/Configuration/TsReference/Index.rst` has no content — the sections below it are not indented into it — so it produces an empty table in the HTML too. Worth removing or filling there, independently of this fix.

Resolves #1435

Assisted-by: Claude Opus 5
Agent-Session: https://claude.ai/code/session_01RxZ2DkBeChDfVViLxvazbZ
Agent-Host: 0493f0
Signed-off-by: Sebastian Mendel <github@sebastianmendel.de>


linawolf (approved):
LGTM

PR #1448, opened 2026-09-19
[BUGFIX] Write the full class name of :php-short: into Markdown

linawolf:
In HTML, :php-short: shows the short class name, and the popup
behind it tells the reader the fully qualified name. Markdown has no
popup, so the namespace was lost, and an agent reading the Markdown
could not tell which class was meant or write the use statement.

The role had the fully qualified name for every class, but passed it
on only when the TYPO3 API knew the class. It now passes it on for
PSR, Fluid, example and unknown classes too, and Markdown writes it
instead of the short name. Their popups gain the name and the use
statement as well.

Assisted-by: Claude Opus 5 <noreply@anthropic.com>
Signed-off-by: lina.wolf


linawolf:
Closing: this change is already on `main` as 176548d07, merged together with #1449, which was built on top of it.

